PR #160, opened 2024-10-11
feat: release

Dav-14:


Earthfile
```diff
@@ -234,3 +242,4 @@ release:
     ARG mode=local
```
flemzord, line 242: Add ARG for VERSION

Earthfile
```diff
@@ -234,3 +242,4 @@ release:
...
     COPY --dir . /src
     DO core+GORELEASER --mode=$mode
+    BUILD +helm-publish
```
flemzord, line 245: Use helm-update ?

Earthfile
Dav-14: already declare and passed implicitly

flemzord (approved):

PR #217, opened 2025-01-14
feat: start ledger v2.2 even in migration not terminated

gfyrag:
the ledger is able to block requests if the schema is no up to date

### DIFF
--- a/internal/resources/ledgers/init.go
+++ b/internal/resources/ledgers/init.go
@@ -93,10 +93,19 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, ledger *v1beta1.Ledger, versio
 	}
 
 	if isV2 && databases.GetSavedModuleVersion(database) != version {
-		if err := migrate(ctx, stack, ledger, database, image, version); err != nil {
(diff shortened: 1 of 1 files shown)


internal/resources/ledgers/init.go
```diff
@@ -93,10 +93,19 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, ledger *v1beta1.Ledger, versio
 	}
 
 	if isV2 && databases.GetSavedModuleVersion(database) != version {
-		if err := migrate(ctx, stack, ledger, database, image, version); err != nil {
+		err := migrate(ctx, stack, ledger, database, image, version)
+		if err != nil {
+			isV2_2 := !semver.IsValid(version) || semver.Compare(version, "v2.2.0-alpha.1") > 0
```
paul-nicolas, line 98: coderabbit may have found a bug, do you want to include `v2.2.0-alpha.1` ?

gfyrag: Yes I want

gfyrag: fixed

PR #239, opened 2025-02-26
feat(ledger): add deployment of the worker

gfyrag:
Fixes ENG-1648

### DIFF
--- a/internal/resources/ledgers/deployments.go
+++ b/internal/resources/ledgers/deployments.go
@@ -2,6 +2,7 @@ package ledgers
 
 import (
 	"fmt"
+	"github.com/formancehq/operator/internal/resources/auths"
 	"golang.org/x/mod/semver"
 	"strconv"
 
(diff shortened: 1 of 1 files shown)


internal/resources/ledgers/deployments.go
```diff
@@ -254,6 +262,43 @@ func installLedgerStateless(ctx core.Context, stack *v1beta1.Stack,
...
+	err := setCommonContainerConfiguration(ctx, stack, ledger, version, database, &container, true)
+	if err != nil {
+		return err
+	}
+
+	//serviceAccountName, err := settings.GetAWSServiceAccount(ctx, stack.Name)
+	//if err != nil {
+	//	return err
+	//}
+
```
Dav-14, line 281: no psql conn in the worker ?

Dav-14 (approved):
LGTM, the operator start being a mess

internal/resources/ledgers/deployments.go
gfyrag: No it's an error. I've uncommented.

PR #297, opened 2025-07-03
fix: namespace for dev envs

gfyrag:


Earthfile
```diff
@@ -71,7 +71,7 @@ deploy:
     END
 
     ARG --required REPOSITORY=
-    RUN --no-cache helm upgrade --install --namespace formance --install formance-operator \
+    RUN --no-cache helm upgrade --install --namespace formance-system --install formance-operator \
```
Dav-14, line 74: looks better :)

Dav-14: It will break current deployment

Dav-14 (approved):

PR #354, opened 2025-11-11
chore: re-order imports using gci

sylr:
Signed-off-by: Sylvain Rabot <sylvain@formance.com>


sylr:
Yes, I'll do that in the upgrade to go 1.25 PR as goalngci-lint needs to be upgraded.

PR #436, opened 2026-04-07
feat: add static settings catalog exporter

altitude:
## Why
The current settings reference is documented in Markdown, which is not a reliable source for deterministic generation of up-to-date docs.

This PR adds a code-derived exporter that scans the operator's Go call sites to `settings.*` and emits a machine-readable catalog from what the operator actually consumes.

## What Changed
- add `cmd/settings-catalog`, a static-analysis based exporter over Go AST call sites
- generate `docs/09-Configuration reference/settings.catalog.json`
- add a focused test for the exporter in `cmd/settings-catalog/main_test.go`
- wire regeneration into `just generate-settings-catalog`
- document the new generation flow in the README

## Output
The generated JSON catalog currently captures:
- setting keys and key patterns
- inferred value types such as `string`, `bool`, `int`, `uri`, `map[string]string`, and `object`
- defaults where the code uses `...OrDefault(...)`
- object field shapes for `GetAs[T]`
- source locations for each discovered setting consumer

## Approach
This is intentionally code-derived using statical analysis.

The exporter walks Go files, finds calls into the `settings` package, and reconstructs key patterns from the arguments passed to those helpers. It also expands helper patterns such as:
- `GetEnvVars(...)` to `...env-vars`
- `GetResourceRequirements(...)` to `...limits`, `...requests`, and `...claims`
- `GetAs[T](...)` to an `object` entry with the inferred object field shape when available

Static analysis was chosen here instead of introducing a runtime registry because the registry does not exist today, and adding one would turn this into a broader refactor of the settings layer rather than a focused exporter.

More concretely:
- the current source of truth for what settings exist is the set of `settings.*` call sites spread across the codebase
- static analysis lets us extract a useful catalog from that existing reality without changing runtime behavior
- it keeps the scope limited to documentation/export concerns rather than mixing in a new metadata model for settings
- it is a good bootstrap step even if we later decide to add a first-class registry and generate both docs and exports from it

In other words, this PR optimizes for extracting deterministic data from the current codebase with minimal behavioral risk. A registry is still a reasonable future direction, but it is a separate design change.

## Current Limits
This does not try to solve everything yet.

In particular, it does not currently provide:
- curated human descriptions or examples
- full URI query parameter semantics
- a first-class registry of settings metadata in code

Those can be layered on later. The goal here is to establish a deterministic machine-readable export from real code usage.

## Validation
- `go test ./cmd/settings-catalog`
- `go run ./cmd/settings-catalog --format json --output "./docs/09-Configuration reference/settings.catalog.json"`

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -4,7 +4,7 @@ set positional-arguments
 default:
   @just --list
 
-pre-commit: tidy lint generate manifests helm-update helm-validate generate-docs
+pre-commit: tidy lint generate manifests helm-update helm-validate generate-docs generate-settings-catalog
 pc: pre-commit
 
 lint:
@@ -98,6 +98,9 @@ generate-docs:
     --output-path="./docs/09-Configuration reference/02-Custom Resource Definitions.md" \
     --templates-dir=./crd-doc-templates \
     --config=./docs.config.yaml
+
+generate-settings-catalog:
+  go run ./cmd/settings-catalog --format json --output "./docs/09-Configuration reference/settings.catalog.json"
 
 deploy: helm-update
   earthly +deploy
--- a/README.md
+++ b/README.md
@@ -205,6 +205,14 @@ make undeploy
 
 See defined settings in [Settings](docs/09-Configuration%20reference/01-Settings.md)
 
+For a machine-readable catalog that can be used by provisioning tooling, run:
+
+```sh
+just generate-settings-catalog
+```
+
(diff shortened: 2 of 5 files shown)


gfyrag (commented):
Imho, overkill, for somethings which just need to be documented when added.

gfyrag (approved):
I don't like it, but no time for that.
And this does not change the runtime of the operator so...

PR #445, opened 2026-04-29
chore(deps): bump go-libs to pick up migrator index name fix

Dav-14:
## Summary
- Bumps `github.com/formancehq/go-libs/v2` from `v2.2.3` to the head of [formancehq/go-libs#595](https://github.com/formancehq/go-libs/pull/595) (`9d6efc1eb666dcc8ceea39c40a7352bf0cea0364`).
- PR #595 is the backport of #594 to `release/v2.2` and fixes a hardcoded `idx_version_id` name collision between migrators sharing a schema (services were crashing at startup with `SQLSTATE 42P10`).
- Affected DBs self-heal on next deploy.

## Notes
- This uses a pseudo-version targeting the PR head. It should be re-pinned to a real `v2.2.x` tag once #595 is merged and tagged.

## Test plan
- [x] `go mod tidy` clean
- [x] `go build ./...` succeeds
- [ ] CI green
- [ ] Smoke deploy verifies migrator no longer panics on startup

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.25
 toolchain go1.25.5
 
 require (
-	github.com/formancehq/go-libs/v2 v2.2.3
+	github.com/formancehq/go-libs/v2 v2.2.5-0.20260429152130-9d6efc1eb666
 	github.com/go-logr/logr v1.4.3
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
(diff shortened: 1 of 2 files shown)


Dav-14:
Closing — the operator does not use the go-libs migrator package, so this bump is unnecessary. The fix only matters for services that actually run migrations.

PR #447, opened 2026-05-05
fix(orchestrations): provision broker topic for events

flemzord:
## Summary
- create a BrokerTopic for orchestration's own published events
- wait for that topic to be ready before deploying orchestration
- reconcile Broker resources when BrokerTopic resources change, so newly-created topics get their NATS streams

## Context
A production stack using OneStreamByService had publisher mapping `*:xjqfdgkwngtt-rzbk-orchestration` with NATS auto-provision disabled, but no matching BrokerTopic/JetStream stream was created. Workflow startup events then failed with `nats: no response from stream`.

## Tests
- `go test ./internal/resources/brokers ./internal/resources/brokertopics ./internal/resources/orchestrations ./internal/resources/brokerconsumers`
- `ASSETS=$(go run sigs.k8s.io/controller-runtime/tools/setup-envtest@latest use 1.32.0 -p path) && CGO_ENABLED=0 KUBEBUILDER_ASSETS="$ASSETS" go test ./internal/tests -ginkgo.focus 'OrchestrationController|BrokerTopicController|BrokerConsumer'`

### DIFF
--- a/internal/resources/brokers/init.go
+++ b/internal/resources/brokers/init.go
@@ -2,6 +2,8 @@ package brokers
 
 import (
 	v1 "k8s.io/api/batch/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/formancehq/operator/v3/api/formance.com/v1beta1"
 	"github.com/formancehq/operator/v3/internal/core"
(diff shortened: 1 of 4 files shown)


fguery (dismissed):

internal/resources/brokers/init.go
```diff
@@ -17,6 +19,13 @@ func init() {
 			core.WithFinalizer[*v1beta1.Broker]("clear", deleteBroker),
 			core.WithOwn[*v1beta1.Broker](&v1.Job{}),
 			core.WithWatchSettings[*v1beta1.Broker](),
+			core.WithWatch[*v1beta1.Broker, *v1beta1.BrokerTopic](func(ctx core.Context, topic *v1beta1.BrokerTopic) []reconcile.Request {
```
gfyrag, line 24: What is the purpose of this change?

flemzord: This watch wakes up the Broker reconciler when a BrokerTopic is created or updated.

In one-stream-by-service mode, the Broker reconciler is responsible for listing BrokerTopic objects and creating the matching NATS stream. When a new topic is introduced after the Broker already exists, there may be no Broker event otherwise, so the stream creation can be delayed until some unrelated reconciliation happens.

I also added a short code comment locally to make this dependency explicit.

gfyrag (approved):

PR #449, opened 2026-05-06
feat: add OtelExporterEndpoint crd

sofiasimdianova:


api/formance.com/v1beta1/otelexporterendpoint_types.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+type OtelSignalConfig struct {
+	// Endpoint URL (e.g., "http://my-collector:4318")
+	Endpoint string `json:"endpoint"`
+
+	// +optional
+	// +kubebuilder:default="http/protobuf"
+	// +kubebuilder:validation:Enum="http/protobuf";"grpc"
+	Protocol string `json:"protocol,omitempty"`
```
Dav-14: can't the protocol be guessed from the endpoint

grpc:// and http://

api/formance.com/v1beta1/otelexporterendpoint_types.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+type SupportConfig struct {
+	Enabled bool `json:"enabled"`
+	// +optional
+	LicenseSecret string `json:"licenseSecret,omitempty"`
+}
```
Dav-14: The licence is already passed with the operator config on startup and available globally within core.Platform

api/formance.com/v1beta1/otelexporterendpoint_types.go
```diff
@@ -0,0 +1,110 @@
...
+	StackDependency `json:",inline"`
+
+	// +optional
+	Traces *OtelSignalConfig `json:"traces,omitempty"`
+
+	// +optional
+	Metrics *OtelSignalConfig `json:"metrics,omitempty"`
+
+	// +optional
+	Support *SupportConfig `json:"support,omitempty"`
```
Dav-14: Can't it be set another way ?

With StackDependecy it's set on only one stack ? Compare to Settings which can be set on multiple stacks.

Can't 1 exporter be set on a stack subset ? or globally ? Using labels ?

cmd/main.go
```diff
@@ -114,10 +116,11 @@ func main() {
...
 	platform := core.Platform{
-		Region:        region,
-		Environment:   env,
-		LicenceSecret: licenceSecret,
-		UtilsVersion:  utilsVersion,
+		Region:          region,
+		Environment:     env,
+		LicenceSecret:   licenceSecret,
+		UtilsVersion:    utilsVersion,
+		SupportEndpoint: supportEndpoint,
```
Dav-14: Can't the support endpoint be used within the CRD itself ?

config/samples/formance.com_v1beta1_otelexporterendpoint.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: formance.com/v1beta1
+kind: OtelExporterEndpoint
+metadata:
+  name: sample-otel
+spec:
+  stack: sample-stack
```
Dav-14: Can stack be a selector ? or a wildcard ?

config/samples/formance.com_v1beta1_otelexporterendpoint.yaml
```diff
@@ -0,0 +1,11 @@
...
+  traces:
+    endpoint: "http://my-collector:4318"
+  support:
+    enabled: true
+    licenseSecret: "formance-license"
```
Dav-14: At the end, 

If we already have the licence-secret on startup we don't need to set it ?

Do we want to enbale the support per-stack ? or per-plane ?

From my POV, the licence is the authentification JWT. and the CRD only need to path the authentification as `BEARER <licence-jwt>`.

The OtelExporterEndpoint could be agnostic of the support.

Creating a support CRD like

```
apiVersion: formance.com/v1beta1
kind: OtelExporterEndpoint
metadata:
  name: sample-otel
spec:
  stackSelector:
    matchLabels:
      formance.com/stack: rrprwuaxwcpy-aarr
    # or
    matchExpressions:
      - key: formance.com/stack
        operator: Exists                 # "any stack"
      - key: formance.com/stack
        operator: In
        values: [rrprwuaxwcpy-aarr, foo]
      - key: formance.com/env
        operator: NotIn
        values: [prod]
  traces:
    endpoint: "http://my-collector:4318"
  auth:
    type: bearer
    fromSecret: "formance-license"

**Comment by Dav-14** [comment]
Dirty is not passing. Make sure to run `just pc`

File: config/samples/formance.com_v1beta1_otelexporterendpoint.yaml (hunk 6 above)
Comment (Dav-14) [reply config/samples/formance.com_v1beta1_otelexporterendpoint.yaml]: Using Label and Expression we lead to some priority problem. In Settings there is something for "*" or specific List that should be enought i guess

File: helm/operator/templates/otelexporterendpoint-support.yaml [hunk 7]
```diff
@@ -0,0 +1,39 @@
...
+kind: OtelExporterEndpoint
+metadata:
+  name: formance-support
+  labels:
+    {{- include "operator.labels" . | nindent 4 }}
+spec:
+  stackSelector:
+    matchExpressions:
+      - key: formance.com/stack
+        operator: Exists
```
Dav-14: can the selector be as it is by default. But exposed in the chart. So that client can modify the selector on need ?

sofiasimdianova: sure

helm/operator/values.yaml
```diff
@@ -86,5 +89,17 @@ affinity: {}
...
+    # Enable the Formance support OtelExporterEndpoint (targets all stacks)
+    enabled: true
+    # The support collector endpoint
+    endpoint: "https://otel-licence.telemetry-support-c1f3ad1.v2.formance.dev"
+    # Secret name containing the bearer token for auth (key: "token").
+    # Defaults to the licence secret if set.
+    authSecret: ""
+    # Resource attributes added to all signals sent to this endpoint
+    resourceAttributes: {}
+
```
Dav-14: Would this shape be reusable accross deployment and helm ? we already populate `global.monitoring.<traces|metrics|logs>` accross `core` lib.

```
global:
  monitoring:
    collector:
        image:
          repository: ""
          tag: ""
        # addtional config in operator case it's managed by him, in other case it could be managed by helm
    support:
        # Enable the Formance support OtelExporterEndpoint (targets all stacks)
        enabled: true
        # The support collector endpoint
        endpoint: "https://otel-licence.telemetry-support-c1f3ad1.v2.formance.dev"
        # Secret name containing the bearer token for auth (key: "token").
        # Defaults to the licence secret if set.
        authSecret: ""
        # Resource attributes added to all signals sent to this endpoint
        resourceAttributes: {}

operator: <----- legacy bad key, totomatic to `config`
```

helm/operator/values.yaml
```diff
@@ -86,5 +89,17 @@ affinity: {}
 webhooks:
   enabled: false
 
+otelExporterEndpoint:
+  support:
+    # Enable the Formance support OtelExporterEndpoint (targets all stacks)
+    enabled: true
```
Dav-14: Can the enabled flag be false by default ?

flemzord (commented):
Inline review comments focused on production-readiness risks, ownership/cleanup semantics, cross-controller reconciliation, validation gaps, and failure modes.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,470 @@
...
+	for i := range stacks.Items {
+		if err := reconcileStackCollector(ctx, &stacks.Items[i]); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func selectorFromSpec(ls *metav1.LabelSelector) (labels.Selector, error) {
+	if ls == nil {
```
flemzord, line 264: A missing `stackSelector` currently matches every stack. For a cluster-scoped resource this is a risky default, and it also contradicts the standard Kubernetes `LabelSelector` description that a null selector matches no objects. Please make this explicit: either require `stackSelector`, treat nil as match-none and `{}` as match-all, or document/default the match-all behavior very visibly.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,470 @@
...
+}
+
+func Cleanup(ctx Context, endpoint *v1beta1.OtelExporterEndpoint) error {
+	selector, err := selectorFromSpec(endpoint.Spec.StackSelector)
+	if err != nil {
+		return err
+	}
+
+	var stacks v1beta1.StackList
+	if err := ctx.GetClient().List(ctx, &stacks, client.MatchingLabelsSelector{Selector: selector}); err != nil {
```
flemzord, line 225: This finalizer only reconciles stacks matching the endpoint's current selector. If the selector was changed before deletion, or a stack label changed earlier, namespaces that used to match are never revisited, so their collector can remain orphaned. Please reconcile the previous `status.stacks` set as well, or otherwise compute all previously/currently affected stacks.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,470 @@
...
+	return metav1.LabelSelectorAsSelector(ls)
+}
+
+func reconcileStackCollector(ctx Context, stack *v1beta1.Stack) error {
+	endpoints, err := findMatchingEndpoints(ctx, stack)
+	if err != nil {
+		return err
+	}
+
+	if len(endpoints) == 0 {
```
flemzord: A matching endpoint is considered active even when it has neither `traces` nor `metrics`, or when the signal endpoint is empty. That still creates a collector and the generated config falls back to `nop`, so workloads can start sending telemetry to a collector that silently drops it. The API/controller should reject or ignore endpoints with no real signal destination.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,470 @@
...
+	}
+
+	annotations := map[string]string{
+		"config-hash": HashFromConfigMaps(configMap),
+	}
+	if secretHashes != "" {
+		annotations["secret-hash"] = secretHashes
+	}
+
+	replicas := int32(1)
```
flemzord, line 353: The collector workload is hard-coded as a single replica and the pod spec below does not expose production controls such as resources, probes, security context, imagePullSecrets, nodeSelector/affinity/tolerations, topology spread, or a PDB. For a per-stack telemetry ingress point this is brittle in production. Please make these knobs configurable, at least through Helm values propagated into the controller or through the CRD/policy.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,470 @@
...
+			}
+			return nil
+		},
+		WithOwner[*corev1.Service](ctx.GetScheme(), stack),
+	)
+	if err != nil {
+		return fmt.Errorf("creating collector service: %w", err)
+	}
+
+	return nil
```
flemzord: Returning nil here lets the standard wrapper mark the `OtelExporterEndpoint` Ready immediately after creating the objects. This does not verify that the Deployment is available, that the collector config starts successfully, or that referenced Secrets contain the expected `token` key. Please add a readiness condition tied to the collector Deployment/status and validate the auth Secret shape before reporting Ready.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,470 @@
...
+func buildCollectorInputs(endpoints []v1beta1.OtelExporterEndpoint) ([]collectorInput, []corev1.EnvVar) {
+	var inputs []collectorInput
+	var envVars []corev1.EnvVar
+
+	refs := referencedAuthSecrets(endpoints)
+	refsByKey := map[string]string{}
+	for _, ref := range refs {
+		envName := fmt.Sprintf("AUTH_%s_%s", envSafe(ref.CRDName), ref.Signal)
+		refsByKey[ref.CRDName+"/"+ref.Signal] = envName
+		envVars = append(envVars, EnvFromSecret(envName, ref.SecretName, "token"))
```
flemzord: This reads the auth Secret from each stack namespace, but the `OtelExporterEndpoint` is cluster-scoped and only carries a Secret name. Existing operator patterns replicate such shared Secrets via `ResourceReference`; this controller does not create one or watch Secrets/ResourceReferences. The Helm support endpoint defaults to the license Secret name, which is rendered in the operator release namespace, so support mode can fail with `auth secret ... not found` until some unrelated module happens to copy it.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,470 @@
...
+	return input, nil
+}
+
+func cleanupStackCollector(ctx Context, namespace string) error {
+	if err := DeleteIfExists[*corev1.Service](ctx, types.NamespacedName{
+		Name: serviceName, Namespace: namespace,
+	}); err != nil {
+		return err
+	}
+	if err := DeleteIfExists[*appsv1.Deployment](ctx, types.NamespacedName{
```
flemzord: High risk: cleanup deletes fixed-name `otel-collector` resources without checking that they are managed by this feature. If a stack namespace already has a user-provided collector Service/Deployment/ConfigMap with these names, removing or retargeting an endpoint can delete it. Please add a stable management label/annotation or owner marker and only mutate/delete resources carrying that marker.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,470 @@
...
+		WithStdReconciler(Reconcile,
+			WithFinalizer[*v1beta1.OtelExporterEndpoint](collectorFinalizer, Cleanup),
+			WithWatch[*v1beta1.OtelExporterEndpoint, *v1beta1.Stack](func(ctx Context, _ *v1beta1.Stack) []reconcile.Request {
+				return enqueueAllEndpoints(ctx)
+			}),
+			WithWatch[*v1beta1.OtelExporterEndpoint, *v1beta1.Settings](func(ctx Context, _ *v1beta1.Settings) []reconcile.Request {
+				return enqueueAllEndpoints(ctx)
+			}),
+			WithRaw[*v1beta1.OtelExporterEndpoint](func(ctx Context, b *builder.Builder) error {
+				collectorPredicate := predicate.NewPredicateFuncs(isCollectorResource)
```
flemzord, line 930: This predicate matches any ConfigMap/Deployment/Service named like the collector anywhere in the cluster, then requeues every endpoint. Combined with the unfiltered Stack and Settings watches above, unrelated objects or non-OTEL Settings can trigger a global endpoints x stacks recalculation. Please filter on a management marker and on the specific OTEL Settings keys, and use the event context for the relist.

internal/resources/otelexporterendpoints/collector_config.go
```diff
@@ -0,0 +1,304 @@
...
+func groupByProcessor(contributions []pipelineContribution) map[string][]string {
+	grouped := map[string][]string{}
+	for _, c := range contributions {
+		grouped[c.processor] = append(grouped[c.processor], c.exporter)
+	}
+	return grouped
+}
+
+func sanitizeName(name string) string {
+	return strings.ReplaceAll(name, ".", "-")
```
flemzord: This can collide for distinct valid Kubernetes names. For example, `support.prod` and `support-prod` both become `support-prod`; `envSafe` also maps both to `SUPPORT_PROD`. That can silently overwrite exporters/processors/env vars and route auth to the wrong destination. Please add a collision-proof suffix, such as a short hash of the original CRD name, or detect collisions and fail clearly.

internal/resources/settings/opentelemetry.go
```diff
@@ -15,9 +17,20 @@ type MonitoringType string
...
+	collectorServicePort = 4318
 )
 
-func GetOTELEnvVars(ctx core.Context, stack, serviceName string, sliceStringSeparator string) ([]v1.EnvVar, error) {
+func GetOTELEnvVars(ctx core.Context, stack, serviceName string, sliceStringSeparator string) ([]corev1.EnvVar, error) {
+	collectorEndpoint, err := getCollectorEndpoint(ctx, stack)
+	if err != nil {
+		return nil, err
+	}
+	if collectorEndpoint != "" {
```
flemzord: Creating or deleting the collector Service changes the env vars produced for every module using `GetOTELEnvVars`, but this controller only requeues `OtelExporterEndpoint` objects. Existing module Deployments will not necessarily roll out when the collector appears, disappears, or changes. Please add a signal that reconciles the stack's dependent workloads when collector presence/config changes.

internal/resources/settings/opentelemetry.go
```diff
@@ -34,7 +47,77 @@ func GetOTELEnvVars(ctx core.Context, stack, serviceName string, sliceStringSepa
...
+	slices.Sort(resourceAttributesArray)
+
+	return []corev1.EnvVar{
+		core.Env("OTEL_TRACES", "true"),
+		core.Env("OTEL_TRACES_BATCH", "true"),
+		core.Env("OTEL_TRACES_EXPORTER", "otlp"),
+		core.Env("OTEL_TRACES_EXPORTER_OTLP_ENDPOINT", collectorEndpoint),
+		core.Env("OTEL_TRACES_EXPORTER_OTLP_MODE", "http"),
+		core.EnvFromBool("OTEL_TRACES_EXPORTER_OTLP_INSECURE", true),
+		core.Env("OTEL_METRICS", "true"),
```
flemzord: When any collector Service exists, this path always enables both traces and metrics for workloads. However the collector config uses a `nop` exporter for a missing signal, so a traces-only endpoint enables metrics and drops them, and a metrics-only endpoint enables traces. Please make activation signal-aware so workloads only emit signals with at least one real destination.

api/formance.com/v1beta1/otelexporterendpoint_types.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+type OtelAuthConfig struct {
+	// +kubebuilder:validation:Enum=bearer
+	Type       string `json:"type"`
+	FromSecret string `json:"fromSecret"`
+}
+
+type OtelSignalConfig struct {
+	Endpoint string `json:"endpoint"`
```
flemzord, line 50: The CRD currently requires `endpoint` only by presence, so an empty string is valid. `fromSecret` has the same issue, and the spec also allows neither `traces` nor `metrics`. Please add validation such as `MinLength=1`, a CEL rule requiring at least one signal, and optionally allowed schemes (`http`, `https`, `grpc`) so invalid specs fail at admission instead of creating collectors that drop data or crash.

internal/resources/otelexporterendpoints/init.go
sofiasimdianova: the nop fallback is intentional (RFC Phase 1) and ensures the collector config remains valid when a signal has no destination. But I agree we should not create a collector when no real signal destinations are configured at all (neither traces nor metrics). I’ll add a check to skip such endpoints to avoid deploying a collector that would drop all telemetry

internal/resources/otelexporterendpoints/init.go
sofiasimdianova: it's deferred in the rfc as Unresolved Questions, that's why in phase 1 replicas and resource configuration are intentionally fixed to keep the crd focused on connectivity concerns (endpoint, auth, selector etc), while operational tuning is handled separately via the existing settings mechanism
this separation is intentional to avoid mixing deployment-level configuration (replicas, resources, probes, scheduling) with the crd responsibility. These knobs will be added in future through settings keys such as otelexporterendpoint.collector.replicas and resource configurations as explained in the rfc

internal/resources/otelexporterendpoints/init.go
sofiasimdianova: for the default use case it's true, the licence secret is already handled by the existing operator mechanism (it gets replicated to each stack namespace via ResourceReference during licence reconciliation, so the formance-licence secret is expected to be present automatically)
but cause rfc also states that a missing secret in one stack should not block other stacks so I'll fix that 

but what I don't understand how it need to be ideally and who actually owns secret lifecycle in the system because right now this controller just assumes secrets already exist (either from ResourceReference or user-managed) and only consumes them

flemzord (commented):
Second deep review pass after the latest updates. I focused on production readiness and remaining behavioral risks.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,631 @@
...
+	inputs, envVars := buildCollectorInputs(activeEndpoints)
+
+	hasTraces, hasMetrics := computeActiveSignals(activeEndpoints, otelSettings)
+
+	collectorConfigYAML, err := generateMergedCollectorConfig(inputs, otelSettings)
+	if err != nil {
+		return fmt.Errorf("generating collector config: %w", err)
+	}
+
+	configMap, _, err := CreateOrUpdate(ctx, types.NamespacedName{
```
flemzord, line 322: This still uses fixed collector names with `CreateOrUpdate`. If a stack namespace already contains a user-managed `otel-collector-config`, this mutator overwrites `Data` and the later `WithLabels` call marks it as managed, making it eligible for future cleanup. Please preflight existing ConfigMap/Deployment/Service objects and refuse to adopt them unless they already have the expected owner or managed label.

helm/operator/templates/otelexporterendpoint-support.yaml
```diff
@@ -0,0 +1,37 @@
...
+    {{- include "operator.labels" . | nindent 4 }}
+spec:
+  stackSelector:
+    {{- toYaml .Values.global.monitoring.support.stackSelector | nindent 4 }}
+  traces:
+    endpoint: {{ .Values.global.monitoring.support.endpoint | quote }}
+    {{- if $authSecret }}
+    auth:
+      type: bearer
+      fromSecret: {{ $authSecret | quote }}
```
flemzord, line 24: This rendered secret name is not enough for a cluster-scoped `OtelExporterEndpoint`: the controller reads `fromSecret` from each stack namespace. The chart creates the licence secret in the Helm release namespace, so support monitoring can fail for stacks where a `ResourceReference` has not copied it. Please create/drive a `ResourceReference` for this auth secret, or make the API carry source namespace/name and replicate it here.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,631 @@
...
+
+		secret := &corev1.Secret{}
+		err := ctx.GetClient().Get(ctx, types.NamespacedName{
+			Name:      ref.SecretName,
+			Namespace: stackNamespace,
+		}, secret)
+		if err != nil {
+			return "", fmt.Errorf("auth secret %q in namespace %q: %w", ref.SecretName, stackNamespace, err)
+		}
+		if err := json.NewEncoder(digest).Encode(secret.Data); err != nil {
```
flemzord: While reading the auth Secret for the rollout hash, please also validate the required `token` key. Today a Secret can exist without `token`, reconciliation proceeds to creating Pods, and the user only gets a generic collector readiness failure at runtime instead of an actionable status error.

internal/resources/settings/opentelemetry.go
```diff
@@ -15,9 +17,20 @@ type MonitoringType string
...
+	collectorServicePort = 4318
 )
 
-func GetOTELEnvVars(ctx core.Context, stack, serviceName string, sliceStringSeparator string) ([]v1.EnvVar, error) {
+func GetOTELEnvVars(ctx core.Context, stack, serviceName string, sliceStringSeparator string) ([]corev1.EnvVar, error) {
+	info, err := getCollectorInfo(ctx, stack)
+	if err != nil {
+		return nil, err
+	}
+	if info != nil {
```
flemzord, line 30: Switching to collector env vars is a behavior change for every workload caller of `GetOTELEnvVars`, but `OtelExporterEndpoint` reconciliation only requeues endpoints. Creating/removing the collector, or flipping the trace/metric Service annotations, will not necessarily reconcile existing Deployments, so they can keep stale OTEL env vars until another event rolls them. Please enqueue affected stack workloads or add a shared rollout signal when collector presence/signals change.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,631 @@
...
+					MatchLabels: map[string]string{
+						"app.kubernetes.io/name": deploymentName,
+					},
+				},
+				Template: corev1.PodTemplateSpec{
+					ObjectMeta: metav1.ObjectMeta{
+						Labels:      collectorLabels(),
+						Annotations: podAnnotations,
+					},
+					Spec: corev1.PodSpec{
```
flemzord, line 371: This pod spec is still hard-coded for the collector. Before production, we need at least resource requests/limits, probes, securityContext/podSecurityContext, imagePullSecrets/scheduling knobs, and ideally a PDB/replica policy. Otherwise enabling this CRD creates per-stack production workloads with no operator/Helm way to set the usual runtime safeguards.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,631 @@
...
+}
+
+func init() {
+	Init(
+		WithStdReconciler(Reconcile,
+			WithFinalizer[*v1beta1.OtelExporterEndpoint](collectorFinalizer, Cleanup),
+			WithWatch[*v1beta1.OtelExporterEndpoint, *v1beta1.Stack](func(ctx Context, _ *v1beta1.Stack) []reconcile.Request {
+				return enqueueAllEndpoints(ctx)
+			}),
+			WithWatch[*v1beta1.OtelExporterEndpoint, *v1beta1.Settings](func(ctx Context, _ *v1beta1.Settings) []reconcile.Request {
```
flemzord: This watch requeues every `OtelExporterEndpoint` for any `Settings` change, even unrelated modules/settings. On large clusters this becomes a global endpoints x stacks scan. Please filter to the OpenTelemetry keys read by `readOtelSettings`/`collectorEnvVars`, or index endpoints/stacks so unrelated Settings updates do not fan out.

docs/09-Configuration reference/02-Custom Resource Definitions.md
```diff
@@ -2177,6 +2178,95 @@ GatewayHTTPAPI is the Schema for the HTTPAPIs API
...
+
+
+
+
+
+| Field | Description | Default | Validation |
+| --- | --- | --- | --- |
+| `ready` _boolean_ | Ready indicates if the resource is seen as completely reconciled |  |  |
+| `info` _string_ | Info can contain any additional like reconciliation errors |  |  |
+| `stacks` _string array_ |  |  |  |
```
flemzord: `status.stacks` is operationally important now because cleanup uses it to find previously matched stacks. The generated docs should describe that contract, otherwise users will not understand why this status field exists or what to expect during selector changes/deletion.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,863 @@
...
+				Kind:    "Secret",
+			}
+			return nil
+		},
+			WithOwner[*v1beta1.ResourceReference](ctx.GetScheme(), stack),
+		)
+		if err != nil {
+			return fmt.Errorf("creating ResourceReference for secret %q in stack %q: %w", ref.SecretName, stack.Name, err)
+		}
+		if !rr.Status.Ready {
```
flemzord, line 781: The pre-existing Secret path is fine, but this fallback path can still stall. When the Secret is missing, this creates a `ResourceReference` and returns `NewPendingError()`. I do not see an `OtelExporterEndpoint` watch on the created `ResourceReference` or on the replicated Secret, and application errors do not requeue automatically here. Once the RR controller marks this RR Ready, the endpoint may not reconcile again until an unrelated Stack/Settings/endpoint event happens. Please add a watch/requeue for these RR/Secret updates and cover the fallback with an envtest that starts from a source Secret labelled `formance.com/stack=any`.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,863 @@
...
+		WithStdReconciler(Reconcile,
+			WithFinalizer[*v1beta1.OtelExporterEndpoint](collectorFinalizer, Cleanup),
+			WithWatch[*v1beta1.OtelExporterEndpoint, *v1beta1.Stack](func(ctx Context, _ *v1beta1.Stack) []reconcile.Request {
+				return enqueueAllEndpoints(ctx)
+			}),
+			WithRaw[*v1beta1.OtelExporterEndpoint](func(ctx Context, b *builder.Builder) error {
+				b.Watches(&v1beta1.Settings{}, handler.EnqueueRequestsFromMapFunc(
+					func(_ context.Context, obj client.Object) []reconcile.Request {
+						s := obj.(*v1beta1.Settings)
+						if !strings.HasPrefix(s.Spec.Key, "opentelemetry.") {
```
flemzord: This predicate fixes the previous global fan-out, but it drops valid wildcard Settings keys. The Settings resolver accepts wildcards per segment, so keys like `*.traces.dsn`, `opentelemetry.*.dsn`, or `*.metrics.resource-attributes` can affect the OTEL values read by this controller without starting with `opentelemetry.`. Please match the supported OTEL key patterns using the same segment/wildcard semantics instead of a string prefix.

api/formance.com/v1beta1/otelexporterendpoint_types.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// OtelSignalConfig configures a single signal type (traces or metrics).
+// Each signal type has its own endpoint and authentication block, allowing
+// different destinations or credentials per signal.
+// Protocol is inferred from the URL scheme: grpc:// for gRPC, http:// or https:// for HTTP/protobuf (default).
+type OtelSignalConfig struct {
+	// Endpoint URL for the signal (e.g., "http://my-collector:4318", "grpc://my-collector:4317").
+	// Protocol is inferred from the URL scheme. HTTP/protobuf is the default for firewall compatibility.
+	// +kubebuilder:validation:MinLength=1
```
flemzord, line 48: `MinLength` still allows unsupported or ambiguous endpoint schemes. The config builder treats only `grpc://` as gRPC and sends everything else through the HTTP exporter, so values like `ftp://collector`, typos, or bare `host:port` can be accepted at admission and fail later in the collector. Please validate allowed schemes (`http`, `https`, `grpc`) or explicitly normalize/document bare host:port before generating the collector config.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,863 @@
...
+	}
+
+	podAnnotations := map[string]string{
+		"config-hash": HashFromConfigMaps(configMap),
+	}
+	if secretHashes != "" {
+		podAnnotations["secret-hash"] = secretHashes
+	}
+
+	replicas := int32(1)
```
flemzord, line 353: The added resources/probes/securityContext are a useful improvement, but the collector runtime policy is still hard-coded here: one replica, fixed requests/limits, fixed probes/securityContext, and no scheduling/PDB/serviceAccount/imagePullSecret knobs. If this is intentionally phase-one behavior, please document that limitation clearly; otherwise expose a collector policy through Helm/platform settings before relying on this in production.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,863 @@
...
+
+	if len(stackErrors) > 0 {
+		endpoint.SetError(fmt.Sprintf("errors in stacks: %s", strings.Join(stackErrors, "; ")))
+	} else {
+		endpoint.SetError("")
+	}
+
+	condition := v1beta1.NewCondition("CollectorsReady", endpoint.GetGeneration())
+	if len(pendingStacks) > 0 {
+		condition.Fail(fmt.Sprintf("waiting for collectors in: %s", strings.Join(pendingStacks, ", ")))
```
flemzord, line 152: This condition now waits for Deployment availability, which is good, but the message remains hard to act on when the collector fails to start. For config errors, image pull failures, CrashLoopBackOff, or failed health checks, users will only see the stack name here and need to inspect Pods manually. Consider surfacing the relevant Deployment condition or latest Pod waiting reason in `status.info`/the condition message.

docs/09-Configuration reference/02-Custom Resource Definitions.md
```diff
@@ -2177,6 +2178,95 @@ GatewayHTTPAPI is the Schema for the HTTPAPIs API
...
+
+
+
+
+
+
+| Field | Description | Default | Validation |
+| --- | --- | --- | --- |
+| `stackSelector` _[LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#labelselector-v1-meta)_ | StackSelector is a standard Kubernetes LabelSelector (matchLabels/matchExpressions).<br />One CRD can target all current and future stacks with a single selector.<br />Matches the pattern established by Settings. |  |  |
+| `traces` _[OtelSignalConfig](#otelsignalconfig)_ | Traces configures the traces signal. At least one of traces or metrics must be set.<br />Logs are intentionally out of scope. |  |  |
```
flemzord, line 2412: These links still point to `#otelsignalconfig`, but the generated document does not define an `OtelSignalConfig` section. `status.stacks` is now documented, so that part is fixed; this remaining doc issue is specifically the missing `OtelSignalConfig`/`OtelAuthConfig` generated sections or broken anchors.

flemzord (commented):
Follow-up review after the latest force-push. I resolved obsolete threads and opened focused comments for the remaining current risks.

api/formance.com/v1beta1/otelexporterendpoint_types.go
```diff
@@ -0,0 +1,124 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// OtelAuthConfig configures per-signal authentication.
+// Auth is per-signal so traces and metrics can use different credentials if needed.
+type OtelAuthConfig struct {
+	// Type is the authentication type.
+	// +kubebuilder:validation:Enum=bearer
+	Type string `json:"type"`
+	// FromSecret references a Secret name (expected key is "token").
```
sylr: You should not hardcode the key and should add a `FromSecretKey` property.

internal/resources/otelexporterendpoints/init.go
```diff
@@ -0,0 +1,904 @@
...
+		}
+
+		if reason := checkCollectorReady(ctx, stack.Name); reason != "" {
+			pendingStacks = append(pendingStacks, fmt.Sprintf("%s (%s)", stack.Name, reason))
+		}
+	}
+
+	for _, prev := range endpoint.Status.Stacks {
+		if !targetedStacks[prev] {
+			if err := cleanupStackCollector(ctx, prev); err != nil {
```
flemzord: [P2] When this endpoint stops targeting a stack, this deletes the stack collector unconditionally. If another OtelExporterEndpoint or the legacy OTel Settings fallback still targets that stack, changing this endpoint selector or a stack label removes the collector and its ResourceReferences until another endpoint happens to reconcile. Mirror the finalizer path here: load the previous stack and call `reconcileStackCollector`, falling back to cleanup only when the stack no longer exists or no signals remain.

internal/resources/otelexporterendpoints/collector_config.go
```diff
@@ -0,0 +1,347 @@
...
+	if protocol == "grpc" && isInsecure(input.signal.Endpoint) {
+		tls = &otlpTLSConfig{Insecure: true}
+	}
+
+	prefix := "otlphttp/"
+	if protocol == "grpc" {
+		prefix = "otlp/"
+	}
+	return prefix + input.name, otlpExporter{
+		Endpoint: endpoint,
```
flemzord, line 143: [P2] The CRD documents bare `host:port` as HTTP, and the sample also uses signal-specific `/v1/traces` and `/v1/metrics` paths, but `buildExporter` always writes the value into the otlphttp generic `endpoint` unchanged. The OpenTelemetry collector expects that field to be a base URL with scheme and appends signal paths itself, so accepted inputs like `my-collector:4318` or copied sample URLs can generate unusable exporter config. Either normalize/reject bare endpoints and document base URLs only, or render signal-specific endpoint fields when paths are supplied.

flemzord (commented):
Inline review comments from Codex.

api/formance.com/v1beta1/otelexporterendpoint_types.go
flemzord: Resolved: `+kubebuilder:validation:Pattern=^(https?://|grpc://)` now rejects bare host:port and unsupported schemes at admission.

internal/resources/otelexporterendpoints/init.go
flemzord: Resolved: the Phase 1 limitation is now documented both in code (comment above the Deployment spec) and in the RFC Unresolved Questions ("Replicas and scheduling"). Configurability via Settings is deferred to a later phase per RFC-0008.

internal/resources/otelexporterendpoints/init.go
flemzord: Resolved: checkCollectorReady now surfaces Deployment Available/Progressing condition messages, and podWaitingReason surfaces container Waiting.Reason/Message (ImagePullBackOff, CrashLoopBackOff, ...) into the CollectorsReady condition.

docs/09-Configuration reference/02-Custom Resource Definitions.md
flemzord: Resolved: the crd-doc-templates/type.tpl fix (operator/v3/api package path + seen-types dedup) now generates the OtelSignalConfig and OtelExporterAuth sections, so the anchors resolve.

internal/resources/otelexporterendpoints/init.go
flemzord: Resolved: Reconcile now mirrors the finalizer path — previously matched stacks are re-reconciled via reconcileStackCollector (which re-evaluates all CRDs and Settings), and direct cleanup only happens when the stack no longer exists.

internal/resources/otelexporterendpoints/collector_config.go
flemzord: Resolved: bare host:port is now rejected at admission by the endpoint pattern, and stripSignalPaths strips /v1/traces and /v1/metrics suffixes before rendering the otlphttp endpoint.

flemzord (approved):
Approving: CI fully green (unit + E2E on K8s 1.31-1.35), all previously raised review threads are addressed and resolved, and the implementation matches RFC-0008 v2 (formancehq/internal-rfcs#13).

One non-blocking follow-up remains open: the envtest for the secret replication fallback path (source Secret labelled `formance.com/stack=any`) — see the open thread on init.go. Fine to land it in a follow-up PR.

PR #460, opened 2026-05-22
fix(benthos): create resource references for envFrom secrets/configmaps

sylr:
Each entry in .spec.envFrom that references a Secret or ConfigMap now gets a ResourceReference so the operator copies the source into the stack namespace (envFrom resources live outside it) and rolls the benthos deployment when their content changes. Stale references from removed/reordered envFrom entries are cleaned up.

Constraint: stack namespace cannot directly reference secrets/configmaps from other namespaces — ResourceReference is the existing copy mechanism
Rejected: hash the user-provided names only | would not roll pods when secret content rotates, defeating the purpose of envFrom
Confidence: high
Scope-risk: narrow
Not-tested: end-to-end roll behavior on secret content change (verified via build only)

### DIFF
--- a/internal/resources/benthos/controller.go
+++ b/internal/resources/benthos/controller.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


internal/resources/benthos/controller.go
```diff
@@ -70,6 +71,66 @@ func createService(ctx Context, b *v1beta1.Benthos) error {
...
+			refName := fmt.Sprintf("envfrom-%d-secret", i)
+			ref, err := resourcereferences.Create(ctx, b, refName, envFrom.SecretRef.Name, &corev1.Secret{})
+			if err != nil {
+				return nil, err
+			}
+			expected[refName] = struct{}{}
+			hashes[refName] = ref.Status.Hash
+		}
+		if envFrom.ConfigMapRef != nil && envFrom.ConfigMapRef.Name != "" {
+			refName := fmt.Sprintf("envfrom-%d-configmap", i)
```
Dav-14: How other ref are named ?

Dav-14: Otherwise LGTM

Dav-14 (approved):

PR #476, opened 2026-06-08
fix(core): normalize partial-semver versions + drop dead e2e carve-out

Dav-14:
## Summary

`ValidateMinimumVersion("v3.2")` silently returned `nil` because `semver.IsValid("v3.2")` is **false** (`golang.org/x/mod/semver` requires `v<major>.<minor>.<patch>`). The short-circuit at `semver.IsValid(…) && Compare(…) < 0` meant every partial-semver Versions resource name (`v3`, `v3.2`, …) bypassed the minimum-version gate entirely. If `MinimumStackVersion` is ever bumped past such a name, the gate would silently let it through.

## Fix

- Add `normalizePartialSemver` which expands `v3` → `v3.0.0` and `v3.2` → `v3.2.0` so partial-semver names are gated by the same minimum as their canonical form. Non-matching inputs (canonical semver, dev tags, SHA refs, non-`v`-prefixed strings) are returned unchanged and continue to pass through.

## Drive-by cleanups in the same function

- **`if strings.TrimPrefix(version, "v") == "0.0.0-e2e" { return nil }` → removed.** It was dead. `ValidateMinimumVersion` has exactly one call site (in `GetModuleVersion`) that fires only when `stack.Spec.VersionsFromFile` is set AND both `module.Version` and `stack.Spec.Version` are empty. Every `tests/e2e/chainsaw/**/resources/stack.yaml` uses `spec.version: v0.0.0-e2e`, so the validator never sees the carve-out for them. semver already understands `-e2e` as a valid pre-release suffix anyway — a real e2e tag at or above the minimum sails through naturally.
- **`if !strings.HasPrefix(version, "v") { version = "v" + version }` → removed.** `golang.org/x/mod/semver` contractually requires the `v` prefix; non-prefixed strings are exactly the "non-semver names allowed through" case the docstring already calls out. `semver.IsValid` returns false for them and they pass through — consistent and removes the only remaining branch in the function body.

## Test plan

- [x] `go test ./internal/core/... -run TestValidateMinimumVersion` — 17/17 pass
- [x] `go build ./internal/core/…` clean
- [x] `go vet ./internal/core/…` clean

### Test-table delta

| Removed | Why |
|---|---|
| `2.1.0 rejected without v prefix` | Hinged on the now-removed prefix-injection branch |
| `2.2.0 accepted without v prefix` | Same — replaced by explicit `non-v-prefixed accepted as passthrough` |
| `v0.0.0-e2e accepted as test tag` | Carve-out gone; the call path never reaches this branch for chainsaw stacks |
| `0.0.0-e2e accepted as test tag without v prefix` | Same |

| Added | Why |
|---|---|
| `v3 accepted (expands to v3.0.0)` | Pins the new partial-semver normalization happy path |
| `v3.2 accepted (expands to v3.2.0)` | Pins the case that motivated this PR |
| `v2.2 accepted as equal to minimum (expands to v2.2.0)` | Boundary at the minimum |
| `v2 rejected (expands to v2.0.0)` | Pins below-minimum partial |
| `v2.1 rejected (expands to v2.1.0)` | Pins below-minimum partial with minor |
| `non-v-prefixed accepted as passthrough` | Makes the new contract explicit |

### DIFF
--- a/internal/core/version.go
+++ b/internal/core/version.go
@@ -3,6 +3,7 @@ package core
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"golang.org/x/mod/semver"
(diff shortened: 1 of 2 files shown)


internal/core/version.go
```diff
@@ -19,18 +20,34 @@ var ErrNoVersionFound = errors.New("no version found")
...
+		minor = "0"
 	}
-	if semver.IsValid(normalizedVersion) && semver.Compare(normalizedVersion, MinimumStackVersion) < 0 {
+	return fmt.Sprintf("v%s.%s.0", m[1], minor)
+}
+
+// ValidateMinimumVersion checks that a Versions resource name meets the minimum requirement.
+// Non-semver names (dev tags, SHA refs, non-`v`-prefixed strings) are allowed through.
+func ValidateMinimumVersion(version string) error {
+	normalized := normalizePartialSemver(version)
```
flemzord, line 49: Preserve validation for versions without the v prefix. The previous contract normalized 2.1.0 to v2.1.0 and rejected it below MinimumStackVersion; this patch instead classifies it as an arbitrary tag and removes that regression case. Because versionsFromFile is a Kubernetes object name and does not require a v prefix, an existing resource named 2.1.0 would now bypass the minimum-version guard and allow the unsupported v2.1 stack again. Please normalize the optional prefix independently from partial-component expansion and retain the below-minimum non-prefixed test.

flemzord (changes requested):
Requesting changes because non-prefixed semver values can now bypass the minimum-version guard. Please also update the primary Stack documentation example: docs/04-Modules/01-Stack.md still recommends versionsFromFile: v2.0, which this change will correctly reject against the v2.2.0 minimum.

Dav-14:
Closing without merging — extracted into #508 to track the underlying issue for later pickup.

PR #482, opened 2026-06-24
fix(gateway): map OTLP mode to valid protocol for Caddy

sofiasimdianova:


internal/resources/caddy/caddy.go
```diff
@@ -37,15 +37,35 @@ func DeploymentTemplate(
...
+			}
+			tracesEndpoint := fmt.Sprintf("%s://$(OTEL_TRACES_EXPORTER_OTLP_ENDPOINT)", scheme)
+			if strings.Contains(endpoint, "://") {
+				tracesEndpoint = "$(OTEL_TRACES_EXPORTER_OTLP_ENDPOINT)"
+			}
+			if protocol == "http/protobuf" {
+				tracesEndpoint += "/v1/traces"
+			}
+			env = append(env, core.Env("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", tracesEndpoint))
+			env = append(env, core.Env("OTEL_EXPORTER_OTLP_PROTOCOL", protocol))
```
NumaryBot: 🟠 **[major] Update the chainsaw protocol expectation**

With the existing observability chainsaw input (`grpc://otel-collector.chainsaw:4317?insecure=true`), this now renders `OTEL_EXPORTER_OTLP_PROTOCOL=grpc`, but `tests/e2e/chainsaw/09-observability/chainsaw-test.yaml` still greps for `OTEL_EXPORTER_OTLP_PROTOCOL=$(OTEL_TRACES_EXPORTER_OTLP_MODE)`. That e2e test will fail whenever the suite is run unless the expectation is updated with this behavior change.

NumaryBot (commented):
## 🛑 Changes requested — automated review

The code change leaves an existing e2e test assertion expecting the previous protocol value, so the repository's test suite can fail even though the runtime mapping may be intended.

internal/resources/caddy/caddy.go
```diff
@@ -132,6 +123,44 @@ func ComputeCaddyfile(ctx core.Context, stack *v1beta1.Stack, _tpl string, addit
...
+
+	protocol := mode
+	if protocol == "http" || protocol == "https" {
+		protocol = "http/protobuf"
+	}
+	tracesEndpoint := fmt.Sprintf("%s://$(OTEL_TRACES_EXPORTER_OTLP_ENDPOINT)", scheme)
+	if strings.Contains(endpoint, "://") {
+		tracesEndpoint = "$(OTEL_TRACES_EXPORTER_OTLP_ENDPOINT)"
+	}
+	if protocol == "http/protobuf" && !strings.HasSuffix(tracesEndpoint, "/v1/traces") {
```
NumaryBot: 🟠 **[major] Check the real endpoint path before appending**

When `opentelemetry.traces.dsn` includes a path, `OTEL_TRACES_EXPORTER_OTLP_ENDPOINT` is the full URL, but this check runs against the literal string `$(OTEL_TRACES_EXPORTER_OTLP_ENDPOINT)`. For HTTP DSNs that already point at a signal endpoint, e.g. `http://collector:4318/v1/traces`, this always appends another suffix and Caddy receives `.../v1/traces/v1/traces`, so traces are exported to the wrong URL.

NumaryBot (commented):
## 🛑 Changes requested — automated review

The new protocol mapping fixes the basic HTTP mode case, but pathful HTTP OTLP DSNs that already include the traces endpoint are malformed by the added suffix logic.

NumaryBot (commented):
## 🛑 Changes requested — automated review

The protocol mapping is mostly correct, but the new endpoint rewriting breaks supported OTLP DSNs that already specify a custom/full trace path.

internal/resources/caddy/caddy.go
```diff
@@ -132,6 +123,44 @@ func ComputeCaddyfile(ctx core.Context, stack *v1beta1.Stack, _tpl string, addit
...
+
+	protocol := mode
+	if protocol == "http" || protocol == "https" {
+		protocol = "http/protobuf"
+	}
+	tracesEndpoint := fmt.Sprintf("%s://$(OTEL_TRACES_EXPORTER_OTLP_ENDPOINT)", scheme)
+	if strings.Contains(endpoint, "://") {
+		tracesEndpoint = "$(OTEL_TRACES_EXPORTER_OTLP_ENDPOINT)"
+	}
+	if protocol == "http/protobuf" && !strings.HasSuffix(endpoint, "/v1/traces") {
```
NumaryBot: 🟠 **[major] Preserve custom OTLP trace paths**

When `opentelemetry.traces.dsn` includes a path, `otelEnvVars` intentionally keeps the full URL in `OTEL_TRACES_EXPORTER_OTLP_ENDPOINT`; this append rewrites that user-specified endpoint to add `/v1/traces` (and if the URL has a query such as `?insecure=true`, the suffix is appended after the query, producing an invalid URL). Please only add the default trace path when the configured endpoint has no path.

NumaryBot (commented):
## 🛑 Changes requested — automated review

The protocol mapping is mostly correct, but base HTTP endpoints with a trailing slash are routed to the wrong path and will fail to export traces through Caddy.

internal/resources/caddy/caddy.go
```diff
@@ -132,6 +123,43 @@ func ComputeCaddyfile(ctx core.Context, stack *v1beta1.Stack, _tpl string, addit
...
+	if mode == "" {
+		return nil
+	}
+
+	protocol := mode
+	if protocol == "http" || protocol == "https" {
+		protocol = "http/protobuf"
+	}
+	tracesEndpoint := fmt.Sprintf("%s://$(OTEL_TRACES_EXPORTER_OTLP_ENDPOINT)", scheme)
+	if strings.Contains(endpoint, "://") {
```
NumaryBot, line 152: 🟠 **[major] Treat base HTTP URLs as needing the traces path**

When `opentelemetry.traces.dsn` is configured as a base HTTP URL with a trailing slash, e.g. `http://collector:4318/`, `otelEnvVars` preserves it as a full endpoint because the parsed path is `/`. This branch then skips appending `/v1/traces`, so with `http/protobuf` Caddy will POST traces to `/` instead of the OTLP traces route. Consider treating `/` as no custom path before deciding not to append the signal path.

internal/resources/caddy/caddy.go
```diff
@@ -132,6 +124,48 @@ func ComputeCaddyfile(ctx core.Context, stack *v1beta1.Stack, _tpl string, addit
...
+		tracesEndpoint = "$(OTEL_TRACES_EXPORTER_OTLP_ENDPOINT)"
+		if protocol == "http/protobuf" {
+			if u, err := url.Parse(endpoint); err == nil && (u.Path == "" || u.Path == "/") {
```
NumaryBot, line 155: 🟠 **[major] Preserve root endpoint without producing double slashes**

When the configured OTLP endpoint is a full URL ending in `/` (for example `http://collector:4318/`), `url.Parse` sets `u.Path` to `/`, but `strings.TrimSuffix` is applied to the literal `$(OTEL_TRACES_EXPORTER_OTLP_ENDPOINT)` rather than the actual endpoint value. After Kubernetes env expansion this becomes `http://collector:4318//v1/traces`, which many OTLP HTTP receivers will not match as `/v1/traces`.

NumaryBot (commented):
## 🛑 Changes requested — automated review

The new protocol mapping generally addresses the intended Caddy value, but it mishandles full HTTP endpoints that include a trailing slash by generating a double-slash OTLP traces path.

tests/e2e/chainsaw/09-observability/chainsaw-test.yaml
```diff
@@ -41,3 +43,29 @@ spec:
...
             content: ../../scripts/dump-diagnostics.sh observability-gateway-traces
+    - name: gateway-opentelemetry-traces-otelexporterendpoint
+      try:
+        - apply:
+            file: resources/otelexporterendpoint.yaml
+        - script:
+            timeout: 2m
+            content: |
+              set -eu
+              kubectl wait --for=jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="OTEL_TRACES_EXPORTER_OTLP_MODE")].value}'=http deployment/gateway -n chainsaw-observability --timeout=2m
```
NumaryBot, line 56: 🔴 **[blocker] Reconcile the gateway before waiting for collector env**

In this second step the only change is applying the `OtelExporterEndpoint`, but the gateway controller watches Settings/Stack generation/owned resources, not `OtelExporterEndpoint` or the managed `otel-collector` Service, so an already-created gateway deployment is not reconciled when the collector appears. In this scenario the deployment still has the first step's `OTEL_TRACES_EXPORTER_OTLP_MODE=grpc`, so this wait for `=http` times out unless the test creates the endpoint before the gateway or explicitly triggers a gateway reconcile.

NumaryBot (commented):
## 🛑 Changes requested — automated review

The new e2e step expects the gateway deployment to update after only an OtelExporterEndpoint is applied, but no gateway reconcile is triggered by that resource, so the test will time out.

NumaryBot (commented):
## ✅ Approve — automated review

No discrete correctness issues were identified in the changed code. The OTLP mode mapping and endpoint path handling appear consistent with the intended Caddy/OpenTelemetry protocol values.

_No findings._

NumaryBot (commented):
## ✅ Approve — automated review

The OTLP protocol mapping and endpoint path handling appear consistent with the existing environment variable generation and the added test coverage. I did not find any discrete regressions introduced by this patch.

_No findings._

flemzord (approved):

PR #484, opened 2026-07-02
fix: encode postgres secret credentials in database uri

### LINKED ISSUE #481: Incorrect handling of passwords in K8s secrets containing characters using special characters
Hello,

I noticed that we got an issue when have a password in a secret for our database using ^. 

When trying to find the cause for it, it seems there should be a url.QueryEscape(password) on this line,  https://github.com/formancehq/operator/blob/2eeb87ef23599517026272b4ae94f035746d6320/internal/resources/databases/env.go#L33.

Compare this with line 27, https://github.com/formancehq/operator/blob/2eeb87ef23599517026272b4ae94f035746d6320/internal/resources/databases/env.go#L27, where it's being encoded.

/Markus

flemzord:
## Summary
- create an encoded PostgreSQL credentials Secret when database credentials come from a Kubernetes Secret
- use encoded username/password env vars when building `POSTGRES_URI`
- keep `POSTGRES_USERNAME` and `POSTGRES_PASSWORD` sourced from the original Secret for existing consumers

## Root cause
Database credentials loaded through `?secret=...` were injected raw and then interpolated into `postgresql://user:password@host:port`, so special URI characters such as `^` could break the resulting connection string.

Fixes #481

## Validation
- `CGO_ENABLED=0 go test ./internal/resources/databases`
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.32.0-darwin-arm64" CGO_ENABLED=0 go test ./...`
- `nix develop -c just pc`

### DIFF
--- a/internal/resources/databases/secret.go
+++ b/internal/resources/databases/secret.go
@@ -1,0 +1,63 @@
+package databases
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/formancehq/operator/v3/api/formance.com/v1beta1"
+	"github.com/formancehq/operator/v3/internal/core"
+)
+
+const (
+	postgresCredentialsUsernameKey = "username"
+	postgresCredentialsPasswordKey = "password"
+
+	encodedPostgresCredentialsSecretSuffix = "postgres-uri-credentials"
+)
+
+func getEncodedPostgresCredentialsSecretName(database *v1beta1.Database) string {
+	return fmt.Sprintf("%s-%s", database.Name, encodedPostgresCredentialsSecretSuffix)
+}
+
+func reconcileEncodedPostgresCredentialsSecret(ctx core.Context, stack *v1beta1.Stack, database *v1beta1.Database, secretName string) error {
+	sourceSecret := &corev1.Secret{}
+	if err := ctx.GetClient().Get(ctx, types.NamespacedName{
+		Namespace: stack.Name,
+		Name:      secretName,
+	}, sourceSecret); err != nil {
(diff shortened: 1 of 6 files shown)


internal/resources/databases/secret.go
```diff
@@ -0,0 +1,64 @@
...
+		return fmt.Errorf("postgres credentials secret %s/%s is missing %q", stack.Name, secretName, postgresCredentialsPasswordKey)
+	}
+
+	_, _, err := core.CreateOrUpdate[*corev1.Secret](ctx, types.NamespacedName{
+		Namespace: stack.Name,
+		Name:      getEncodedPostgresCredentialsSecretName(database),
+	}, func(secret *corev1.Secret) error {
+		secret.Type = corev1.SecretTypeOpaque
+		secret.Data = map[string][]byte{
+			postgresCredentialsUsernameKey: []byte(url.QueryEscape(string(username))),
```
NumaryBot: 🟠 **[major] url.QueryEscape encodes spaces as '+' which is invalid in URI userinfo**

When a Secret username or password contains a space, url.QueryEscape encodes it as '+'. However, '+' is not interpreted as a space in the URI userinfo section (it is only decoded as a space in query strings). This means credentials containing spaces will be embedded incorrectly in the database URI, causing authentication failures with the wrong decoded value.

**Suggestion:** Use url.PathEscape (or url.QueryEscape followed by replacing '+' with '%20') for encoding credentials placed in the userinfo portion of the URI, so that spaces are emitted as '%20' rather than '+'.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/484#issuecomment-4864312472

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/484#issuecomment-4864312472

sofiasimdianova (approved):

PR #487, opened 2026-07-07
fix(operator): avoid encoded postgres secret collision

flemzord:
## Summary
- avoid overwriting a user-provided Postgres credentials Secret when its name collides with the derived encoded credentials Secret name
- keep the existing derived Secret name for non-conflicting deployments, and use a collision-safe name only when needed
- only delete encoded credentials Secrets controlled by the Database reconciler, so cleanup cannot remove an uncontrolled source Secret
- add regression tests for env rendering, Secret reconciliation idempotence, and cleanup behavior

## Root cause
`v3.12.0` introduced a derived Secret named `<database>-postgres-uri-credentials` for URL-encoded Postgres credentials. A real deployment can already use that exact name as the source Secret referenced by `?secret=...` (for example `formance-ledger-postgres-uri-credentials`). In that case, the Database reconciler writes encoded credentials back into the source Secret it just read, turning its input into its output and causing repeated reconciliations/rollouts around Ledger deployments.

## Helm / release context
`formance-1.15.0` resolves `regions` to `3.12.0`, which resolves the Operator chart to `3.12.0`; that is how the Operator change from #484 reached users. No Helm template change is needed for the fix itself, but a follow-up Operator release and Helm dependency refresh will be needed to ship it through the umbrella chart.

## Validation
- `CGO_ENABLED=0 go test ./internal/resources/databases -run 'Test(GetPostgresEnvVarsAvoidsEncodedSecretNameCollision|ReconcileEncodedPostgresCredentialsSecretAvoidsSourceSecretNameCollision|DeleteEncodedPostgresCredentialsSecretKeepsUncontrolledSourceSecret)' -count=1 -v`
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.32.0-darwin-arm64" CGO_ENABLED=0 go test ./...`
- `nix develop -c just pre-commit`

## Operational checks
- `prod-eu-west-1-hosting`: Operator is still `v3.11.2`; Ledger databases use `?secret=...`, so it is exposed if upgraded to `3.12.0` before this fix.
- `staging-eu-west-1-hosting`: Operator is on the post-`v3.12.0` image, but visible Ledger databases use inline credentials, so it does not exercise the affected `?secret=...` path.
- No matching `object has been modified` / `POSTGRES_URL_ENCODED` / `postgres-uri-credentials` loop was found in the last 24h of our prod/staging hosting operator logs.

### DIFF
--- a/internal/resources/databases/init.go
+++ b/internal/resources/databases/init.go
@@ -61,11 +61,15 @@ func Reconcile(ctx core.Context, stack *v1beta1.Stack, database *v1beta1.Databas
 			return err
 		}
 	} else {
-		err = resourcereferences.Delete(ctx, database, "postgres")
+		previousPostgresCredentialsSecret, err := getPostgresCredentialsSecretReference(ctx, database)
 		if err != nil {
 			return err
 		}
-		if err := deleteEncodedPostgresCredentialsSecret(ctx, stack, database); err != nil {
+		if err := deleteEncodedPostgresCredentialsSecret(ctx, stack, database, previousPostgresCredentialsSecret); err != nil {
+			return err
+		}
+		err = resourcereferences.Delete(ctx, database, "postgres")
+		if err != nil {
 			return err
 		}
 	}
@@ -114,7 +118,20 @@ func Reconcile(ctx core.Context, stack *v1beta1.Stack, database *v1beta1.Databas
 	return nil
 }
 
+func getPostgresCredentialsSecretReference(ctx core.Context, database *v1beta1.Database) (string, error) {
+	reference := &v1beta1.ResourceReference{}
+	if err := ctx.GetClient().Get(ctx, types.NamespacedName{
+		Name: fmt.Sprintf("%s-postgres", database.Name),
+	}, reference); err != nil {
+		return "", client.IgnoreNotFound(err)
+	}
+	return reference.Spec.Name, nil
+}
+
 func Delete(ctx core.Context, database *v1beta1.Database) error {
+	if err := remediatePostgresCredentialsSecretBeforeDatabaseDelete(ctx, database); err != nil {
+		return err
+	}
 	if database.Status.URI == nil {
 		return nil
 	}
@@ -137,6 +154,9 @@ func Delete(ctx core.Context, database *v1beta1.Database) error {
 	}, stack); err != nil {
 		return err
 	}
+	if err := reconcileEncodedPostgresCredentialsSecretBeforeDatabaseDelete(ctx, stack, database); err != nil {
+		return err
+	}
 
 	if err := handleDatabaseJob(ctx, stack, database, "drop-database", "db", "drop"); err != nil {
 		return err
@@ -146,6 +166,49 @@ func Delete(ctx core.Context, database *v1beta1.Database) error {
 	logger.Info("Database deleted.")
 
 	return nil
+}
+
+func remediatePostgresCredentialsSecretBeforeDatabaseDelete(ctx core.Context, database *v1beta1.Database) error {
+	if database.Spec.Stack == "" {
+		return nil
+	}
+	sourceSecretName, err := getPostgresCredentialsSecretReference(ctx, database)
+	if err != nil {
+		return err
+	}
+	sourceSecretNames := make([]string, 0, 2)
+	if sourceSecretName != "" {
+		sourceSecretNames = append(sourceSecretNames, sourceSecretName)
+	}
+	if database.Status.URI != nil {
+		sourceSecretNameFromStatus := database.Status.URI.Query().Get("secret")
+		if sourceSecretNameFromStatus != "" {
+			sourceSecretNames = append(sourceSecretNames, sourceSecretNameFromStatus)
+		}
+	}
+	stack := &v1beta1.Stack{}
+	stack.Name = database.Spec.Stack
+	sourceSecretNames = dedupePostgresCredentialsSecretNames(sourceSecretNames)
+	if len(sourceSecretNames) == 0 {
+		return deleteDefaultEncodedPostgresCredentialsSecretWithoutSource(ctx, stack, database, getEncodedPostgresCredentialsSecretName(database))
+	}
+	for _, sourceSecretName := range sourceSecretNames {
+		if err := removeSourceSecretControllerReferenceByName(ctx, stack, database, sourceSecretName); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func reconcileEncodedPostgresCredentialsSecretBeforeDatabaseDelete(ctx core.Context, stack *v1beta1.Stack, database *v1beta1.Database) error {
+	if database.Status.URI == nil {
+		return nil
+	}
+	sourceSecretName := database.Status.URI.Query().Get("secret")
+	if sourceSecretName == "" {
+		return nil
+	}
+	return reconcileEncodedPostgresCredentialsSecret(ctx, stack, database, sourceSecretName)
 }
 
 func handleDatabaseJob(ctx core.Context, stack *v1beta1.Stack, database *v1beta1.Database, name string, args ...string) error {
--- a/internal/resources/databases/secret.go
+++ b/internal/resources/databases/secret.go
@@ -1,11 +1,16 @@
 package databases
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/formancehq/operator/v3/api/formance.com/v1beta1"
 	"github.com/formancehq/operator/v3/internal/core"
@@ -15,11 +20,48 @@ const (
 	postgresCredentialsUsernameKey = "username"
 	postgresCredentialsPasswordKey = "password"
 
-	encodedPostgresCredentialsSecretSuffix = "postgres-uri-credentials"
+	encodedPostgresCredentialsSecretSuffix              = "postgres-uri-credentials"
+	collisionSafeEncodedPostgresCredentialsSecretSuffix = "encoded-postgres-uri-credentials"
+	encodedPostgresCredentialsSecretAnnotation          = "formance.com/encoded-postgres-credentials-secret"
 )
 
-func getEncodedPostgresCredentialsSecretName(database *v1beta1.Database) string {
-	return fmt.Sprintf("%s-%s", database.Name, encodedPostgresCredentialsSecretSuffix)
+func getEncodedPostgresCredentialsSecretName(database *v1beta1.Database, sourceSecretName ...string) string {
+	name := fmt.Sprintf("%s-%s", database.Name, encodedPostgresCredentialsSecretSuffix)
+	if len(sourceSecretName) > 0 && sourceSecretName[0] == name {
+		return fmt.Sprintf("%s-%s-%s", database.Name, collisionSafeEncodedPostgresCredentialsSecretSuffix, hashPostgresCredentialsSecretName(sourceSecretName[0]))
+	}
+	return name
+}
+
+func hashPostgresCredentialsSecretName(secretName string) string {
+	hash := sha256.Sum256([]byte(secretName))
+	return hex.EncodeToString(hash[:])[:8]
+}
+
+func getEncodedPostgresCredentialsSecretCandidateNames(database *v1beta1.Database, sourceSecretName ...string) []string {
+	defaultName := fmt.Sprintf("%s-%s", database.Name, encodedPostgresCredentialsSecretSuffix)
+	names := []string{
+		defaultName,
+		fmt.Sprintf("%s-%s", database.Name, collisionSafeEncodedPostgresCredentialsSecretSuffix),
+		getEncodedPostgresCredentialsSecretName(database, defaultName),
+	}
+	if len(sourceSecretName) > 0 && sourceSecretName[0] != "" {
+		names = append(names, getEncodedPostgresCredentialsSecretName(database, sourceSecretName[0]))
+	}
+	return dedupePostgresCredentialsSecretNames(names)
+}
+
+func dedupePostgresCredentialsSecretNames(names []string) []string {
+	ret := make([]string, 0, len(names))
+	seen := make(map[string]struct{}, len(names))
+	for _, name := range names {
+		if _, ok := seen[name]; ok {
+			continue
+		}
+		seen[name] = struct{}{}
+		ret = append(ret, name)
+	}
+	return ret
 }
 
 func reconcileEncodedPostgresCredentialsSecret(ctx core.Context, stack *v1beta1.Stack, database *v1beta1.Database, secretName string) error {
@@ -29,6 +71,9 @@ func reconcileEncodedPostgresCredentialsSecret(ctx core.Context, stack *v1beta1.
 		Name:      secretName,
 	}, sourceSecret); err != nil {
 		return errors.Wrap(err, "getting postgres credentials secret")
+	}
+	if err := removeSourceSecretControllerReference(ctx, database, sourceSecret); err != nil {
+		return err
 	}
 
 	username, ok := sourceSecret.Data[postgresCredentialsUsernameKey]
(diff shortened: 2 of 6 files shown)


internal/resources/databases/secret.go
```diff
@@ -51,13 +66,44 @@ func reconcileEncodedPostgresCredentialsSecret(ctx core.Context, stack *v1beta1.
...
+		if err := deleteEncodedPostgresCredentialsSecretIfControlled(ctx, stack, database, name); err != nil {
+			return errors.Wrap(err, "deleting encoded postgres credentials secret")
+		}
+	}
+	return nil
+}
+
+func deleteStaleEncodedPostgresCredentialsSecrets(ctx core.Context, stack *v1beta1.Stack, database *v1beta1.Database, sourceSecretName, desiredSecretName string) error {
+	for _, name := range getEncodedPostgresCredentialsSecretCandidateNames(database) {
+		if name == sourceSecretName || name == desiredSecretName {
```
NumaryBot, line 204: 🔴 **[blocker] Stale controller ownerReference on colliding source Secrets not cleaned up**

When upgrading a cluster that already experienced the secret name collision, the previous reconciler set the Database as the controller owner of the user-provided source Secret. The current fix skips processing the colliding source Secret, but leaves the stale ownerReference in place. Kubernetes GC and the `Owns(&v1.Secret{})` watch still treat the source credentials Secret as Database-owned, meaning deleting or recreating the Database can garbage-collect the user's credentials despite the collision guard.

**Suggestion:** In addition to skipping the colliding source Secret during normal reconciliation, add a remediation path that detects and removes the stale Database controller ownerReference from the source Secret. This ensures existing clusters that already hit the collision are fully protected after upgrade.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/487#issuecomment-4912311921

sylr (dismissed):

flemzord:
Addressed in the latest push (643a6868):

- remediates already-affected clusters by removing the stale Database controller ownerReference from the source Postgres credentials Secret during reconciliation;
- keeps the user-provided source Secret data intact while creating the collision-safe encoded Secret;
- adds regression coverage for that upgrade/remediation path;
- adds coverage for stale controlled encoded Secret cleanup when the source Secret later changes back to a non-colliding name.

Validation run locally:

- `CGO_ENABLED=0 go test ./internal/resources/databases -count=1`
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.32.0-darwin-arm64" CGO_ENABLED=0 go test ./...`
- `nix develop -c just pre-commit`

internal/resources/databases/secret.go
```diff
@@ -51,13 +70,58 @@ func reconcileEncodedPostgresCredentialsSecret(ctx core.Context, stack *v1beta1.
...
-	})
-	return errors.Wrap(err, "deleting encoded postgres credentials secret")
+		Name:      name,
+	}, secret); err != nil {
+		return client.IgnoreNotFound(err)
+	}
+	if !metav1.IsControlledBy(secret, database) {
+		return nil
+	}
+	return ctx.GetClient().Delete(ctx, secret)
```
NumaryBot, line 225: 🔴 **[blocker] Stale controller ownerReference on colliding source Secrets not cleaned up**

On clusters that already experienced the secret name collision, the previous reconciler set the Database resource as the controller owner of the user-provided source Secret. The current fix skips processing the colliding source Secret during normal reconciliation but does not remove the stale ownerReference. Kubernetes garbage collection and the `Owns(&v1.Secret{})` watch still treat the source credentials Secret as Database-owned. Furthermore, when the Postgres URI is changed to no longer include `?secret=...`, `deleteEncodedPostgresCredentialsSecret` can be entered without first calling any remediation, and the user-provided source Secret (already marked as controlled by this Database) will satisfy `IsControlledBy` and be deleted — destroying user credentials instead of cleaning up stale metadata.

**Suggestion:** Add a remediation step that detects and removes the stale Database controller ownerReference from source Secrets before any delete logic runs. Specifically: (1) during reconciliation, when a source Secret name collides with the encoded Secret name, check whether the source Secret has a stale Database controller ownerReference and remove it; (2) in `deleteEncodedPostgresCredentialsSecret`, before deleting a Secret that satisfies `IsControlledBy`, verify that the Secret is not a user-provided source credential (i.e. its name is not one that would collide) — if it is, remove the ownerReference and skip deletion instead.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/487#issuecomment-4912311921

flemzord:
Follow-up for the latest NumaryBot finding addressed in 0ae381c9:

- the delete path now reads the previous Postgres ResourceReference before deleting it, so it can identify the former source Secret when the URI changes away from `?secret=...`;
- if that former source Secret is one of the encoded-secret candidate names, the operator removes the stale Database controller ownerReference and preserves the Secret instead of deleting it;
- if the ResourceReference is already missing, the default collision-prone candidate is treated conservatively as a possible user source Secret and remediated instead of deleted;
- controlled operator-created encoded Secrets are still deleted when the previous source Secret is known and non-colliding.

Additional validation run locally:

- targeted delete/reconcile regression tests for source Secret remediation and stale cleanup
- `CGO_ENABLED=0 go test ./internal/resources/databases -count=1`
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.32.0-darwin-arm64" CGO_ENABLED=0 go test ./...`
- `nix develop -c just pre-commit`

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/487#issuecomment-4912311921

internal/resources/databases/secret.go
```diff
@@ -30,6 +45,9 @@ func reconcileEncodedPostgresCredentialsSecret(ctx core.Context, stack *v1beta1.
 	}, sourceSecret); err != nil {
 		return errors.Wrap(err, "getting postgres credentials secret")
 	}
+	if err := removeSourceSecretControllerReference(ctx, database, sourceSecret); err != nil {
```
NumaryBot, line 75: 🟡 **[minor] Remediation of stale ownerReferences skipped during Database deletion path**

The stale-ownerReference remediation runs only during normal reconciliation. If a Database object is already being deleted (finalizer path) on a cluster that was previously affected by the collision bug, the remediation step is never reached. The user-provided source Secret can still carry the stale Database controller ownerReference, allowing Kubernetes GC to delete the user's credentials when the Database object is removed.

**Suggestion:** Run the stale-ownerReference remediation unconditionally — both on the normal reconciliation path and before any finalizer/delete logic executes — so that already-affected clusters are fully protected even when the Database is being deleted immediately after upgrade.

flemzord:
Finalizer follow-up addressed in 7ad375cc:

- `Delete` now remediates stale Database controller ownerReferences on the Postgres credentials source Secret before any early return, including the `clear-database=false` path;
- the remediation uses the existing Postgres ResourceReference when available, falls back to `database.status.uri?secret=...`, and finally treats the default collision-prone candidate conservatively;
- added regression coverage for the Database finalizer path and the ResourceReference-based remediation path.

Validation run locally after this change:

- targeted finalizer/delete remediation tests
- `CGO_ENABLED=0 go test ./internal/resources/databases -count=1`
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.32.0-darwin-arm64" CGO_ENABLED=0 go test ./...`
- `nix develop -c just pre-commit`

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/487#issuecomment-4912311921

internal/resources/databases/secret.go
```diff
@@ -40,9 +58,10 @@ func reconcileEncodedPostgresCredentialsSecret(ctx core.Context, stack *v1beta1.
 		return fmt.Errorf("postgres credentials secret %s/%s is missing %q", stack.Name, secretName, postgresCredentialsPasswordKey)
 	}
 
+	encodedSecretName := getEncodedPostgresCredentialsSecretName(database, secretName)
```
NumaryBot, line 88: 🟠 **[major] Fallback encoded Secret name can collide with and overwrite an unrelated pre-existing Secret**

When the source Secret is named `<database>-postgres-uri-credentials`, the code always derives `<database>-encoded-postgres-uri-credentials` as the collision-safe name. `CreateOrUpdate` will silently overwrite any pre-existing, uncontrolled Secret that already carries that name. This recreates the same collision class the PR is trying to fix — just one suffix level deeper — and can destroy unrelated credentials in the namespace.

**Suggestion:** Before writing to the derived fallback name, check whether a Secret with that name already exists and is not controlled by this Database. If such a Secret exists, choose a distinct name (e.g. incorporate a hash of the source name) or surface an error/event rather than overwriting the existing Secret.

internal/resources/databases/secret.go
NumaryBot, line 75: 🟡 **[minor] Stale ownerReference remediation skipped during Database deletion/finalizer path**

The stale-ownerReference remediation introduced to protect user-provided source Secrets from Kubernetes GC only runs on the normal reconciliation path. If a Database object enters the finalizer/delete path on a cluster that was previously affected by the collision bug before normal reconciliation has run post-upgrade, the stale controller ownerReference is never stripped. Kubernetes GC can then destroy the user's credentials when the Database object is removed. This is a narrower variant of the original blocker and was explicitly called out in PR discussion as a residual concern after the prior blockers were addressed.

**Suggestion:** Run the stale-ownerReference remediation unconditionally — both during normal reconciliation and before any finalizer/delete logic executes — so that already-affected clusters are fully protected even when the Database is deleted immediately after upgrade.

flemzord:
Hashed fallback follow-up addressed in 01e1fd30:

- the collision fallback Secret name now includes a deterministic hash of the source Secret name instead of using only the fixed `<database>-encoded-postgres-uri-credentials` suffix;
- the previous fixed fallback name remains in the candidate list only for stale controlled Secret cleanup;
- reconciliation now refuses to write to an existing encoded Secret target unless that Secret is already controlled by the same Database, preventing overwrites of unrelated user Secrets;
- added regression coverage for preserving an uncontrolled legacy fallback Secret and for rejecting an uncontrolled existing hashed target.

Validation run locally after this change:

- targeted fallback collision/remediation tests
- `CGO_ENABLED=0 go test ./internal/resources/databases -count=1`
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.32.0-darwin-arm64" CGO_ENABLED=0 go test ./...`
- `nix develop -c just pre-commit`

internal/resources/databases/secret.go
```diff
@@ -40,9 +84,13 @@ func reconcileEncodedPostgresCredentialsSecret(ctx core.Context, stack *v1beta1.
...
```
NumaryBot, line 88: 🟠 **[major] Fallback encoded Secret name can collide with and silently overwrite an unrelated pre-existing Secret**

When the source Secret is named `<database>-postgres-uri-credentials`, the code always derives `<database>-encoded-postgres-uri-credentials` as the collision-safe fallback name. `CreateOrUpdate` will silently overwrite any pre-existing, uncontrolled Secret that already carries that derived name. This recreates the same collision class the PR is trying to fix — just one suffix level deeper — and can destroy unrelated credentials in the namespace.

**Suggestion:** Before writing to the derived fallback name, check whether a Secret with that name already exists and is not controlled by this Database. If such a Secret exists, choose a distinct name (e.g. incorporating a hash of the source name) or surface an error/event rather than overwriting the existing Secret.

internal/resources/databases/secret.go
```diff
@@ -15,11 +20,47 @@ const (
...
+	hash := sha256.Sum256([]byte(secretName))
+	return hex.EncodeToString(hash[:])[:8]
+}
+
+func getEncodedPostgresCredentialsSecretCandidateNames(database *v1beta1.Database, sourceSecretName ...string) []string {
+	names := []string{
+		fmt.Sprintf("%s-%s", database.Name, encodedPostgresCredentialsSecretSuffix),
+		fmt.Sprintf("%s-%s", database.Name, collisionSafeEncodedPostgresCredentialsSecretSuffix),
+	}
+	if len(sourceSecretName) > 0 && sourceSecretName[0] != "" {
```
NumaryBot, line 48: 🟡 **[minor] Hashed fallback encoded Secret not included in stale-cleanup candidate list after source name change**

When a Database moves from the colliding source name (`<db>-postgres-uri-credentials`) back to a non-colliding `?secret=...` source, the previously created hashed fallback encoded Secret (produced by `getEncodedPostgresCredentialsSecretName(database, oldSource)`) is not visited by `deleteStaleEncodedPostgresCredentialsSecrets` or the delete cleanup logic. The candidate list only adds that hashed name when the current source is empty or still colliding, so the old controlled Secret containing stale encoded credentials is left behind in the stack namespace.

**Suggestion:** Track the previously used encoded Secret name (e.g. via a status field or annotation) and include it in the stale-cleanup candidate list unconditionally, so it is removed when the source changes back to a non-colliding name.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/487#issuecomment-4912311921

flemzord:
Hashed fallback stale cleanup addressed in e55f8109:

- the stale encoded Secret candidate list now always includes the hashed fallback for the historical colliding source name, even when the current source Secret is non-colliding;
- this cleans up controlled hashed fallback Secrets when a Database moves from `<database>-postgres-uri-credentials` back to another `?secret=...` source;
- extended the stale cleanup regression test to cover both the legacy fixed fallback Secret and the new hashed fallback Secret.

Validation run locally:

- targeted stale hashed fallback cleanup test
- `CGO_ENABLED=0 go test ./internal/resources/databases -count=1`
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.32.0-darwin-arm64" CGO_ENABLED=0 go test ./...`
- `nix develop -c just pre-commit`

internal/resources/databases/secret.go
flemzord: Addressed in e55f8109: the fallback name now uses the deterministic hashed target via `getEncodedPostgresCredentialsSecretName(database, secretName)`, and `ensureEncodedPostgresCredentialsSecretCanBeWritten` rejects any existing target Secret that is not controlled by this Database before `CreateOrUpdate` runs. Regression coverage: `TestReconcileEncodedPostgresCredentialsSecretKeepsUncontrolledLegacyFallbackSecret` and `TestReconcileEncodedPostgresCredentialsSecretRejectsUncontrolledEncodedSecretTarget`.

internal/resources/databases/secret.go
flemzord: Addressed in e55f8109: `getEncodedPostgresCredentialsSecretCandidateNames` now always includes the hashed fallback for the historical default-name collision, even when the current source Secret is non-colliding. The stale cleanup test now creates both the legacy fixed fallback and the hashed fallback and asserts both controlled Secrets are removed.

internal/resources/databases/secret.go
NumaryBot, line 88: 🟠 **[major] Fallback encoded Secret name can collide with and silently overwrite an unrelated pre-existing Secret**

When the source Secret is named `<database>-postgres-uri-credentials`, the code derives `<database>-encoded-postgres-uri-credentials` as the collision-safe fallback name. `CreateOrUpdate` will silently overwrite any pre-existing, uncontrolled Secret that already carries that derived name. This recreates the same collision class the PR is trying to fix — just one suffix level deeper — and can destroy unrelated credentials in the namespace. This thread (Thread 6) is marked as still open (no resolution) in the prior discussion, and the author's fix (01e1fd30) incorporates a hash but the reviewer's open thread indicates the concern may not be fully addressed.

**Suggestion:** Before writing to the derived fallback name, verify that a Secret with that name either does not exist or is already controlled by this Database. If an uncontrolled Secret with that name exists, choose a distinct name (e.g. incorporating a hash of the source name) or surface an error/event rather than overwriting it.

internal/resources/databases/secret.go
```diff
@@ -15,11 +20,47 @@ const (
...
+}
+
+func getEncodedPostgresCredentialsSecretCandidateNames(database *v1beta1.Database, sourceSecretName ...string) []string {
+	defaultName := fmt.Sprintf("%s-%s", database.Name, encodedPostgresCredentialsSecretSuffix)
+	names := []string{
+		defaultName,
+		fmt.Sprintf("%s-%s", database.Name, collisionSafeEncodedPostgresCredentialsSecretSuffix),
+		getEncodedPostgresCredentialsSecretName(database, defaultName),
+	}
+	if len(sourceSecretName) > 0 && sourceSecretName[0] != "" {
```
NumaryBot, line 48: 🟡 **[minor] Hashed fallback encoded Secret not included in stale-cleanup candidate list after source name change**

When a Database transitions from the colliding source name (`<db>-postgres-uri-credentials`) back to a non-colliding `?secret=...` source, the previously created hashed fallback encoded Secret is not visited by the stale-cleanup or delete-cleanup logic. The candidate list only adds that hashed name when the current source is still empty or colliding, so the old controlled Secret containing stale encoded credentials is left behind in the namespace. This thread (Thread 7) is still open in the prior discussion.

**Suggestion:** Track the previously used encoded Secret name (e.g. via a status field or annotation) and include it in the stale-cleanup candidate list unconditionally, so it is removed when the source changes back to a non-colliding name.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/487#issuecomment-4912311921

internal/resources/databases/secret.go
flemzord: Still addressed in f62ad62c: `encodedSecretName` is the hashed fallback for the colliding source name, and `ensureEncodedPostgresCredentialsSecretCanBeWritten` prevents `CreateOrUpdate` from touching an uncontrolled existing target. Covered by `TestReconcileEncodedPostgresCredentialsSecretKeepsUncontrolledLegacyFallbackSecret` and `TestReconcileEncodedPostgresCredentialsSecretRejectsUncontrolledEncodedSecretTarget`.

internal/resources/databases/secret.go
flemzord: Still addressed in f62ad62c: `getEncodedPostgresCredentialsSecretCandidateNames` always includes the hashed fallback for the historical default collision source, so stale controlled hashed fallback Secrets are visited even after moving back to a non-colliding source. Covered by `TestReconcileEncodedPostgresCredentialsSecretDeletesStaleControlledEncodedSecret`.

internal/resources/databases/init.go
```diff
@@ -148,6 +165,25 @@ func Delete(ctx core.Context, database *v1beta1.Database) error {
...
+		return nil
+	}
+	sourceSecretName, err := getPostgresCredentialsSecretReference(ctx, database)
+	if err != nil {
+		return err
+	}
+	if sourceSecretName == "" && database.Status.URI != nil {
+		sourceSecretName = database.Status.URI.Query().Get("secret")
+	}
+	if sourceSecretName == "" {
```
NumaryBot: 🟠 **[major] Encoded credentials Secret can be orphaned when status URI is empty at delete time**

If reconciliation creates the encoded credentials Secret but then fails before `database.Status.URI` is recorded, a subsequent deletion of that Database enters the fallback branch with no ResourceReference and no status URI. The code treats the default encoded Secret name as a possible source Secret and removes the Database controller ownerReference. `Delete` then returns early because `Status.URI` is nil, so Kubernetes GC can no longer clean up the operator-created encoded credentials Secret when the Database object is removed, leaving a stale Secret in the namespace.

**Suggestion:** Before removing the Database controller ownerReference in the fallback path, verify whether the candidate Secret was actually created by the operator (e.g. by checking a dedicated label or annotation set at creation time). If it is an operator-created encoded Secret rather than a user source Secret, retain the ownerReference so GC can clean it up, or explicitly delete it before returning.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/487#issuecomment-4912311921

internal/resources/databases/init.go
```diff
@@ -61,11 +61,15 @@ func Reconcile(ctx core.Context, stack *v1beta1.Stack, database *v1beta1.Databas
...
+		previousPostgresCredentialsSecret, err := getPostgresCredentialsSecretReference(ctx, database)
 		if err != nil {
 			return err
 		}
-		if err := deleteEncodedPostgresCredentialsSecret(ctx, stack, database); err != nil {
+		if err := deleteEncodedPostgresCredentialsSecret(ctx, stack, database, previousPostgresCredentialsSecret); err != nil {
+			return err
+		}
+		err = resourcereferences.Delete(ctx, database, "postgres")
+		if err != nil {
```
NumaryBot, line 72: 🟡 **[minor] ResourceReference deleted before encoded Secret on the delete path, causing resource leak on retry**

In the `else` branch, `resourcereferences.Delete` is called before `deleteEncodedPostgresCredentialsSecret`. If encoded-secret deletion fails and the reconcile retries, `getPostgresCredentialsSecretReference` returns an empty string because the reference was already removed. `deleteEncodedPostgresCredentialsSecret` then cannot identify the former source Secret and may preserve the encoded Secret instead of deleting it, leaking the resource.

**Suggestion:** Swap the order: call `deleteEncodedPostgresCredentialsSecret` first, and only delete the ResourceReference after the encoded Secret has been successfully removed. Both operations are idempotent on retry in this order.

internal/resources/databases/init.go
flemzord: Addressed in fbf8125d: encoded credentials Secrets created by the operator are now annotated with `formance.com/encoded-postgres-credentials-secret=true`; the finalizer fallback without ResourceReference or Status.URI deletes an annotated Secret if it is controlled by the Database, while still only removing the ownerReference from unannotated possible source Secrets. Covered by `TestDeleteRemovesAnnotatedEncodedSecretWhenSourceReferenceAndStatusURIAreMissing` and the existing source-preservation test.

internal/resources/databases/init.go
flemzord: Addressed in f62ad62c/fbf8125d: the no-secret reconcile path now calls `deleteEncodedPostgresCredentialsSecret` before `resourcereferences.Delete`, so a failed Secret deletion keeps the postgres ResourceReference for retry. Covered by `TestReconcileKeepsPostgresResourceReferenceWhenEncodedSecretDeleteFails`.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/487#issuecomment-4912311921

internal/resources/databases/init.go
```diff
@@ -148,6 +165,25 @@ func Delete(ctx core.Context, database *v1beta1.Database) error {
...
+		return nil
+	}
+	sourceSecretName, err := getPostgresCredentialsSecretReference(ctx, database)
+	if err != nil {
+		return err
+	}
+	if sourceSecretName == "" && database.Status.URI != nil {
+		sourceSecretName = database.Status.URI.Query().Get("secret")
+	}
+	stack := &v1beta1.Stack{}
```
NumaryBot, line 189: 🟠 **[major] Operator-created encoded Secret can be orphaned or drop-job can hang when Status.URI is empty at delete time**

Two related scenarios arise when the Database is deleted before Status.URI is recorded: (1) If the encoded credentials Secret was created by the operator but lacks the new annotation (e.g. upgraded clusters), the fallback branch removes the Database controller ownerReference and returns early without deleting the Secret, so Kubernetes GC can no longer clean it up, leaving a stale Secret in the namespace. (2) When clear-database=true, the delete path never creates the hashed fallback Secret unless normal reconciliation has already run, so GetPostgresEnvVars causes the drop job to reference a missing Secret and the finalizer never completes.

**Suggestion:** Before removing the Database controller ownerReference in the fallback path, check whether the candidate Secret was created by the operator (e.g. via a dedicated label/annotation set at creation time). If it is operator-created, retain the ownerReference for GC or explicitly delete it. For the drop-job scenario, ensure the encoded Secret is created (or reconstituted) before the drop job is submitted, even on the delete path.

internal/resources/databases/init.go
NumaryBot, line 72: 🟡 **[minor] ResourceReference deleted before encoded Secret on the delete path — resource leak on retry**

In the else branch, resourcereferences.Delete is called before deleteEncodedPostgresCredentialsSecret. If the encoded-secret deletion fails and reconciliation retries, getPostgresCredentialsSecretReference returns an empty string because the reference was already removed. deleteEncodedPostgresCredentialsSecret then cannot identify the former source Secret and may preserve the encoded Secret instead of deleting it, leaking the resource.

**Suggestion:** Swap the order: call deleteEncodedPostgresCredentialsSecret first, and only delete the ResourceReference after the encoded Secret has been successfully removed. Both operations are idempotent on retry in this order.

internal/resources/databases/init.go
flemzord: Addressed in 84af876e: the finalizer fallback now only deletes operator-marked encoded Secrets (`formance.com/encoded-postgres-credentials-secret=true`) and leaves unannotated default Secrets under their existing ownerReference when there is no ResourceReference or Status.URI evidence. The delete path also calls `reconcileEncodedPostgresCredentialsSecretBeforeDatabaseDelete` before submitting the drop job, so a `Status.URI?secret=...` colliding source reconstitutes the hashed encoded Secret first. Covered by `TestDeleteKeepsUnannotatedDefaultSecretOwnerReferenceWhenSourceReferenceAndStatusURIAreMissing`, `TestDeleteRemovesAnnotatedEncodedSecretWhenSourceReferenceAndStatusURIAreMissing`, and `TestReconcileEncodedPostgresCredentialsSecretBeforeDatabaseDeleteCreatesCollisionSafeSecret`.

internal/resources/databases/init.go
flemzord: Addressed in 84af876e: the no-secret reconcile branch still deletes/remediates encoded Secrets before deleting the postgres ResourceReference. This is covered by `TestReconcileKeepsPostgresResourceReferenceWhenEncodedSecretDeleteFails`, which verifies the ResourceReference remains available when Secret deletion fails and the reconcile retries.

internal/resources/databases/init.go
```diff
@@ -148,6 +168,36 @@ func Delete(ctx core.Context, database *v1beta1.Database) error {
...
 
+func remediatePostgresCredentialsSecretBeforeDatabaseDelete(ctx core.Context, database *v1beta1.Database) error {
+	if database.Spec.Stack == "" {
+		return nil
+	}
+	sourceSecretName, err := getPostgresCredentialsSecretReference(ctx, database)
+	if err != nil {
+		return err
+	}
+	if sourceSecretName == "" && database.Status.URI != nil {
```
NumaryBot: 🟠 **[major] Status-URI source Secret may retain stale ownerReference when ResourceReference and Status.URI disagree at delete time**

When a ResourceReference exists but has already been updated to point at a different Postgres secret while database.Status.URI still references the old colliding source Secret, the fallback remediation path skips the Status.URI secret entirely. On the clear-database=false delete path, Delete returns immediately after this remediation, so the old source Secret can still carry the stale Database controller ownerReference and be garbage-collected when the Database is removed — destroying user credentials.

**Suggestion:** In the delete/finalizer path, remediate both the ResourceReference-identified Secret and the Status.URI-identified Secret (when they differ and the Status.URI Secret matches a known colliding candidate name). Ensure the stale ownerReference is removed from the Status.URI source Secret even when a current ResourceReference already points elsewhere.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/487#issuecomment-4912311921

internal/resources/databases/init.go
flemzord: Addressed in ef34297b: the finalizer remediation now collects both the postgres ResourceReference Secret and the `database.Status.URI?secret=...` Secret, deduplicates them, and removes stale Database ownerReferences from each. Covered by `TestRemediatePostgresCredentialsSecretBeforeDatabaseDeleteUsesResourceReferenceAndStatusURI`.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/487#issuecomment-4912311921

flemzord:
Final follow-up after CI review comments:

- fixed the retry leak by keeping the postgres `ResourceReference` until encoded secret cleanup succeeds;
- made finalizer remediation cover both the postgres `ResourceReference` source secret and the `database.status.uri?secret=...` source secret when they disagree;
- annotated operator-owned encoded postgres secrets and constrained finalizer fallback cleanup so ambiguous unannotated default-name secrets are not mutated without source evidence;
- reconstituted the collision-safe encoded secret before clear-database drop jobs.

Validation:
- `CGO_ENABLED=0 go test ./internal/resources/databases -count=1`
- `KUBEBUILDER_ASSETS="$(pwd)/bin/k8s/1.32.0-darwin-arm64" CGO_ENABLED=0 go test ./...`
- `nix develop -c zsh -lc "just pre-commit"`
- GitHub Actions `Default` run `28931949680` is green across tests, Dirty/pre-commit, and E2E Kubernetes 1.31-1.35.
- CodeRabbit: pass; NumaryBot review: pass / no findings; unresolved review threads: 0.

sylr (approved):

PR #488, opened 2026-07-10
fix(databases): preserve encoded postgres credentials

flemzord:
## Summary

- preserve URL-encoded PostgreSQL credentials from Kubernetes Secrets by default, restoring the behavior expected before operator v3.12.0
- add an explicit `secretCredentialsEncoding=raw` mode for Secrets that contain raw credentials, and reject unsupported values
- expose the selected mode as `POSTGRES_CREDENTIALS_ENCODING` so Kubernetes rolls affected workloads when the behavior changes
- filter the internal encoding parameter from the final connection URI and document the v3.12.0 / Formance 1.15.0 upgrade behavior
- add regression coverage for raw, pre-encoded, and rollout-signal behavior and regenerate the CRD/settings artifacts

## Context

Operator v3.12.0 started URL-encoding every Secret-backed database credential. Deployments that had already encoded special characters for earlier operator versions were therefore double-encoded (`%5E` became `%255E`) after upgrading through Formance Helm 1.15.0.

This keeps the legacy URL-encoded contract as the default because auto-detecting encoded credentials is ambiguous for literal `%XX` passwords. Users who store raw credentials can opt in explicitly with:

```text
postgresql://postgres:5432?secret=postgres&secretCredentialsEncoding=raw
```

References [TS-493](https://formance-team.atlassian.net/browse/TS-493). Related to #481.

## Validation

- `nix --extra-experimental-features "nix-command flakes" develop --impure --command just pre-commit`
- `nix --extra-experimental-features "nix-command flakes" develop --impure --command just tests`

## Risk

Deployments created on v3.12.0 that rely on raw credentials containing URI-special characters must add `secretCredentialsEncoding=raw`. Plain credentials without special characters are unchanged.

[TS-493]: https://formance-team.atlassian.net/browse/TS-493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/resources/databases/env.go
+++ b/internal/resources/databases/env.go
@@ -31,6 +31,10 @@ func GetPostgresEnvVars(ctx core.Context, stack *v1beta1.Stack, database *v1beta
 			)
 		} else {
 			secret := database.Status.URI.Query().Get("secret")
+			credentialsEncoding, err := parsePostgresCredentialsEncoding(database.Status.URI.Query().Get(postgresCredentialsEncodingQueryParam))
+			if err != nil {
+				return nil, err
+			}
 			postgresURIUsernameEnv = "POSTGRES_URL_ENCODED_USERNAME"
 			postgresURIPasswordEnv = "POSTGRES_URL_ENCODED_PASSWORD"
 			encodedSecretName := getEncodedPostgresCredentialsSecretName(database, secret)
@@ -39,6 +43,7 @@ func GetPostgresEnvVars(ctx core.Context, stack *v1beta1.Stack, database *v1beta
 				core.EnvFromSecret("POSTGRES_PASSWORD", secret, postgresCredentialsPasswordKey),
 				core.EnvFromSecret("POSTGRES_URL_ENCODED_USERNAME", encodedSecretName, postgresCredentialsUsernameKey),
 				core.EnvFromSecret("POSTGRES_URL_ENCODED_PASSWORD", encodedSecretName, postgresCredentialsPasswordKey),
+				core.Env("POSTGRES_CREDENTIALS_ENCODING", string(credentialsEncoding)),
 			)
 		}
 		ret = append(ret,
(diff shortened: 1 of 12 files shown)


internal/resources/databases/env.go
```diff
@@ -129,6 +129,7 @@ func BuildPostgresQueryString(rawQuery url.Values) string {
 	delete(params, "disableSSLMode")
 	delete(params, "secret")
 	delete(params, "awsRole")
+	delete(params, postgresCredentialsEncodingQueryParam)
```
NumaryBot, line 138: 🔴 **[blocker] Add a rollout signal when credential encoding changes**

When upgrading from v3.12.0 with already-running Secret-backed deployments, filtering `secretCredentialsEncoding` out leaves the pod/job template identical while only the generated Secret data changes. These credentials are consumed as env vars from Secrets, and Kubernetes does not refresh env-var Secret values in existing Pods, so workloads can keep using the old double-encoded or unencoded credentials until a manual restart. Please include a hash/annotation or another spec change tied to the generated credentials/encoding mode so the affected pods/jobs are recreated.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/488#issuecomment-4933724381

internal/resources/databases/env.go
```diff
@@ -130,6 +130,7 @@ func BuildPostgresQueryString(rawQuery url.Values) string {
...
```
NumaryBot, line 138: 🔴 **[blocker] Secret env-var changes do not trigger a pod rollout**

When upgrading from v3.12.0 with already-running Secret-backed deployments, filtering out `secretCredentialsEncoding` leaves the pod/job template spec unchanged while only the underlying Secret data is updated. Kubernetes does not refresh env-var Secret values in running Pods, so workloads continue using the old (double-encoded or unencoded) credentials until a manual restart is performed.

**Suggestion:** Include a hash or annotation derived from the generated credentials/encoding mode in the pod/job template spec (e.g., as an annotation on the pod template) so that Kubernetes detects a spec change and recreates the affected pods/jobs automatically.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/488#issuecomment-4933724381

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/488#issuecomment-4933724381

gfyrag (approved):

PR #490, opened 2026-07-15
feat(ledger): integrate Ledger v3 operator

flemzord:
## Context

This PR integrates the Ledger v3 Operator into the Formance Operator while preserving the historical Ledger v2 behavior.

The goal is to delegate native provisioning for Ledger versions strictly greater than `v3.0.0-alpha`, allow Ledger v3 to run alongside Ledger v2 in preparation for migrations, and avoid making the Ledger Operator mandatory for stacks that do not use these features.

### Native Ledger v3 provisioning

- For Ledger versions strictly greater than `v3.0.0-alpha`, the Operator creates a `Cluster.ledger.formance.com` instead of the historical Deployments, Jobs, CronJobs, and Database resources.
- The replica count continues to use `deployments.ledger.replicas`. It defaults to `3`, and any positive even value is rounded up to the next odd number to preserve quorum (`2 -> 3`, `4 -> 5`).
- The image, `clusterID`, debug mode, TLS, reserved labels, and `serviceName: ledger` remain enforced by the Operator.
- Resources, image pull secrets, and the service account are overridden only when a corresponding value is provided by Settings or the registry configuration.
- `extraEnv`, Auth, and monitoring are merged field by field so that advanced options not managed by Settings are preserved, including Pyroscope, Flight Recorder, `scopeMapping`, and `anonymousScopes`.
- The `deployments.ledger.topology-spread-constraints` Setting is applied to the v3 Cluster.

### Shared configuration through `LedgerConfiguration`

- Adds the cluster-scoped `LedgerConfiguration.formance.com/v1beta1` CRD.
- `spec.cluster` directly reuses the Ledger Operator's `ClusterSpec`: its descriptions and Kubebuilder validations are expanded into the generated CRD without duplicating the Go type.
- `spec.stacks` follows the Settings convention: `['*']` provides the global configuration, while a configuration explicitly targeting a stack takes precedence.
- The selected configuration is used as the base; values computed from the Stack, Settings, Auth, and registry configuration are then applied according to the precedence rules above.
- Creating, updating, or deleting a configuration reconciles the affected Ledgers. When no configuration applies, the historical v3 behavior is preserved.

### Ledger v3 preview for preparing a v2-to-v3 migration

- The `ledger.v3.preview-version` Setting starts an isolated v3 Cluster while fully preserving the Ledger v2 deployment and its Database.
- `/api/ledger/v3` targets the v3 preview, while `/api/ledger/v2` and the historical routes continue to target v2.
- The `ledger.BucketService` gRPC service is routed to the v3 Cluster.
- Removing the Setting deletes only the preview resources and routes; Ledger v2 continues to operate.
- No mirroring or final switch is triggered automatically. Data replication and customer validation remain explicit migration steps.

### TLS, Auth, and Gateway exposure

- Creates the cert-manager Issuer and Certificate resources required for inter-node TLS and Gateway-to-Ledger-v3 traffic.
- gRPC routes are published only after the Certificate and its Secret are ready.
- Auth configuration is derived from the stack's Auth module while preserving additional fields defined in `LedgerConfiguration`.
- Adds `GatewayGRPCAPI` and the associated Caddy routing alongside the HTTP rules.

### NetworkPolicies

- Allows Raft and gRPC service traffic between replicas of the same v3 Cluster on ports `7777` and `8888`.
- Uses distinct rules for direct and preview Clusters to preserve the existing selectors.
- Allows Ledger v3 to reach Ledger v2 only within its own Stack namespace on port `8080`, supporting mirror mode without cross-stack access.

### Optional dependencies and safeguards

- On startup, the Operator checks that the `Cluster` CRD exists and that it has `get/list/watch/create/update/patch/delete` permissions before registering the Ledger v3 watcher.
- If the CRD or permissions are unavailable, v2 stacks continue to reconcile normally and v3 previews are ignored without crashing the Operator.
- A stack configured directly for v3 remains pending when the Ledger Operator is unavailable: it does not silently fall back to v2 resources.
- A v2-to-v3 transition is blocked while v2 resources exist, and a v3-to-v2 transition is blocked while the direct v3 Cluster exists. The Operator reports a `MigrationRequired` condition instead of implicitly destroying data.
- Preview cleanup remains possible when cert-manager is no longer available; inaccessible cert-manager resources are ignored during cleanup.

## Expected behavior

1. A v2 stack without the preview Setting continues to work without the Ledger Operator installed.
2. A v2 stack with `ledger.v3.preview-version` and the required CRDs runs v2 and v3 in parallel, with HTTP v2/v3 and gRPC v3 routed correctly.
3. The same stack with the Ledger v3 CRDs unavailable keeps v2 running and ignores the preview without a global reconciliation failure.
4. A stack configured with a Ledger v3 version creates a native Cluster and creates neither a Database nor Ledger v2 workloads.
5. A `LedgerConfiguration` explicitly targeting a stack always takes precedence over the wildcard configuration.
6. Operator-reserved values and Settings take precedence over the shared configuration, while unmanaged fields are preserved.
7. Replica counts are always positive and odd; invalid values are rejected explicitly.
8. With NetworkPolicies enabled, v3 nodes can form their cluster and can read only the Ledger v2 instance in their own namespace.
9. Switching Ledger generations never implicitly deletes the other generation's resources.

## Validation

- `nix develop --command just pc`
  - `go mod tidy`, linting, and generation for the affected modules
  - CRD, RBAC, Helm, and documentation generation
  - Helm chart linting and rendering
- `nix develop --command just tests`
  - 11 Ginkgo suites passed
  - 133 controller specs passed
- `CGO_ENABLED=0 go test ./...` with Kubernetes 1.32 envtest assets
- `nix develop --command golangci-lint run --timeout 5m`: 0 issues
- `helm lint ./helm/operator --strict`
- `helm lint ./helm/crds --strict`
- `git diff --check`
- GitHub Actions E2E matrix passed on Kubernetes 1.31, 1.32, 1.33, 1.34, and 1.35

A non-blocking warning remains during `just tests`: the bundled Ginkgo CLI is `2.27.2`, while the Go packages use `2.28.1`.

## Deployment

The updated CRD chart must be deployed before this Operator version. The Ledger Operator and cert-manager are required to run Ledger v3 Clusters with TLS, but they are not required to continue operating v2 stacks.


internal/resources/ledgers/v3_preview.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	setLedgerV3PreviewCondition(ledger, metav1.ConditionTrue, "Running", message)
+	return nil
+}
+
+func ledgerV3HTTPBackendRef(stackName string) v1beta1.GatewayBackendRef {
+	return v1beta1.GatewayBackendRef{
+		Name: "ledger-" + stackName,
+		Port: ledgerV3HTTPPort,
```
NumaryBot: 🟠 **[major] Gateway backend ports hard-coded instead of derived from LedgerConfiguration**

When a LedgerConfiguration overrides the Ledger Cluster service ports (e.g. spec.cluster.service.grpcPort or httpPort), composeLedgerV3ClusterSpec preserves those configured values, but the Gateway backend refs are still hard-coded to 9000/8888. This means the Ledger Operator exposes the Cluster service on the configured ports while the Gateway keeps proxying to the defaults, silently breaking HTTP/gRPC routing.

**Suggestion:** Either enforce the default ports in the Cluster spec so they always match the hard-coded Gateway refs, or derive the Gateway backend port refs from the Cluster spec's desired port values so the two stay in sync.

internal/resources/ledgers/v3.go
```diff
@@ -0,0 +1,491 @@
...
+		setLedgerV3Condition(ledger, metav1.ConditionFalse, "TLSReconcileFailed", err.Error())
+		return err
+	}
+
+	cluster, err := createOrUpdateV3Cluster(ctx, stack, ledger, version, false, tlsCAHash)
+	if err != nil {
+		setLedgerV3Condition(ledger, metav1.ConditionFalse, "ReconcileFailed", err.Error())
+		return err
+	}
+	if !tlsReady {
```
NumaryBot, line 221: 🟠 **[major] Stale gRPC route left in place when TLS becomes unready**

For the direct (non-preview) Ledger v3 path, if a GatewayGRPCAPI was previously created and the Certificate/Secret later becomes unready or is deleted, the reconciler returns pending without cleaning up the existing gRPC route. The Gateway can then continue exposing a backend whose TLS material is no longer valid. The preview path correctly deletes the route in this same TLS-pending scenario, but the non-preview path does not.

**Suggestion:** Mirror the preview path's behaviour: when the TLS precondition is not met (certificate/secret unready or missing), explicitly delete any existing GatewayGRPCAPI resource before returning pending, so the Gateway does not continue routing to an insecure or broken backend.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

internal/resources/stacks/networkpolicies.go
```diff
@@ -96,11 +98,104 @@ func createNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
...
+			selector := directLedgerV3Selector(stack.Name)
+			np.Spec = networkingv1.NetworkPolicySpec{
+				PodSelector: selector,
+				PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
+				Ingress: []networkingv1.NetworkPolicyIngressRule{
+					{
+						From: []networkingv1.NetworkPolicyPeer{
+							{PodSelector: &selector},
+						},
+						Ports: networkPolicyTCPPorts(7777, 8888),
```
NumaryBot: 🟠 **[major] Ledger v3 NetworkPolicy ports not derived from LedgerConfiguration overrides**

When networkpolicies.enabled is set and a LedgerConfiguration overrides spec.cluster.service.raftPort or grpcPort, composeLedgerV3ClusterSpec preserves those port overrides but the NetworkPolicy still only allows the hard-coded defaults (7777 and 8888). In that configuration, Ledger v3 pods can be created with non-default Raft/gRPC ports while ingress between replicas is denied, preventing the cluster from forming or serving gRPC under NetworkPolicies.

**Suggestion:** Derive the NetworkPolicy ingress ports from the same Cluster spec values used in composeLedgerV3ClusterSpec, so the allowed ports always match the actual service ports.

deployment/operator/helpers.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	platforms := make([]string, 0, len(allPlatforms))
+	for _, p := range allPlatforms {
+		if strings.HasSuffix(p, arch) {
+			platforms = append(platforms, p)
+		}
+	}
+	if len(platforms) == 0 {
+		platforms = []string{"linux-" + arch}
+	}
```
NumaryBot, line 106: 🟠 **[major] Default arch filter silently excludes arm64 from multi-arch images**

Defaulting `arch` to `amd64` when it is unset filters out the `linux-arm64` platform entry, so the 'multi-arch' image effectively becomes amd64-only and cannot run on arm64 nodes.

**Suggestion:** Leave the platform list unfiltered when `arch` is not explicitly set; apply the suffix filter only when `arch` is provided by the caller.

deployment/operator/helpers.go
```diff
@@ -0,0 +1,272 @@
...
+						Ref: pulumi.Sprintf("%s/%s:buildcache-%s", dc.Registry, name, platform),
+					},
+				},
+			},
+			CacheTo: dockerbuild.CacheToArray{
+				dockerbuild.CacheToArgs{
+					Registry: dockerbuild.CacheToRegistryArgs{
+						Ref: pulumi.Sprintf("%s/%s:buildcache-%s,mode=max", dc.Registry, name, platform),
+					},
+				},
```
NumaryBot, line 165: 🟠 **[major] Cache mode appended to image Ref string instead of typed field**

Appending `,mode=max` directly to the cache image `Ref` string produces an invalid image reference. The cache mode should be set via the typed `CacheToRegistryArgs.Mode` field.

**Suggestion:** Remove `,mode=max` from the `Ref` string and set `Mode: "max"` on the `CacheToRegistryArgs` struct instead.

deployment/operator/main.go
```diff
@@ -0,0 +1,128 @@
...
+			return fmt.Errorf("failed to build operator image: %w", err)
+		}
+
+		// Apply CRDs
+		crdFiles, err := filepath.Glob(filepath.Join("..", "..", "config", "crd", "bases", "*.yaml"))
+		if err != nil {
+			return fmt.Errorf("failed to glob CRD files: %w", err)
+		}
+
+		var crds []pulumi.Resource
```
NumaryBot, line 45: 🟠 **[major] No error raised when CRD glob returns an empty result**

`filepath.Glob` returns an empty slice without error when no files match. Because Helm values disable chart-managed CRDs, an incorrect working directory or incomplete package would silently deploy the operator without any CRDs installed.

**Suggestion:** After the glob call, add a check: `if len(crdFiles) == 0 { return fmt.Errorf("no CRD files found under config/crd/bases") }`.

deployment/operator/main.go
```diff
@@ -0,0 +1,128 @@
...
+		if licenceToken != "" {
+			licenceIssuer := cfg.Get("licence-issuer")
+			if licenceIssuer == "" {
+				licenceIssuer = "https://license.formance.cloud/keys"
+			}
+			licenceValues = pulumi.Map{
+				"createSecret": pulumi.Bool(true),
+				"token":        pulumi.String(licenceToken),
+				"issuer":       pulumi.String(licenceIssuer),
+			}
```
NumaryBot: 🟠 **[major] licence-token secret value exposed as plaintext in Pulumi state**

`cfg.Get("licence-token")` retrieves the secret config value as plaintext before passing it to `pulumi.String(...)`, stripping Pulumi's secret tracking and potentially leaking the token into state snapshots and logs.

**Suggestion:** Use `config.GetSecret(ctx, "licence-token")` to keep the value encrypted in the Pulumi state graph.

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    fi; \
+    CGO_ENABLED=0 go build -buildvcs=false -o /usr/bin/operator -ldflags="${LDFLAGS}" .
+
+FROM alpine:3.20
+
+RUN apk update && apk add --no-cache ca-certificates curl
+
+ENTRYPOINT ["/usr/bin/operator"]
+
+COPY --from=builder /usr/bin/operator /usr/bin/operator
```
NumaryBot, line 35: 🟠 **[major] Operator container runs as root**

The final image stage does not declare an unprivileged user, so the operator process runs as root. This weakens containment if the operator is ever compromised.

**Suggestion:** Add `RUN addgroup -S operator && adduser -S -G operator operator` after the `apk` install step and add `USER operator` after copying the binary.

internal/resources/gateways/Caddyfile.gotpl
```diff
@@ -133,16 +149,41 @@
 
 				{{- if or (not (semver_is_valid $values.Gateway.Version)) (gt (semver_compare $values.Gateway.Version "v0.1.7") 0) }}
 				{{ $service.Name }} {
-					http://{{ $service.Name }}:8080/_info http://{{ $service.Name }}:8080/{{ $service.HealthCheckEndpoint }}
+					http://{{ $service.HealthCheckBackend.Name }}:{{ $service.HealthCheckBackend.Port }}/_info http://{{ $service.HealthCheckBackend.Name }}:{{ $service.HealthCheckBackend.Port }}/{{ $healthCheckEndpoint }}
 				}
 				{{- else }}
-				{{ $service.Name }} http://{{ $service.Name }}:8080/_info http://{{ $service.Name }}:8080/{{ $service.HealthCheckEndpoint }}
+				{{ $service.Name }} http://{{ $service.HealthCheckBackend.Name }}:{{ $service.HealthCheckBackend.Port }}/_info http://{{ $service.HealthCheckBackend.Name }}:{{ $service.HealthCheckBackend.Port }}/{{ $healthCheckEndpoint }}
```
NumaryBot, line 185: 🟠 **[major] Health-check probe URLs use http:// even for TLS-enabled backends**

The `/versions` health-check probes remain `http://` even when `HealthCheckBackend.TLS` is configured. TLS-only backends will receive plaintext probes and may appear unavailable.

**Suggestion:** Generate TLS-aware probe URLs by emitting `https://` when `HealthCheckBackend.TLS` is set, and apply the backend's configured CA trust settings to these probes.

internal/resources/ledgers/init.go
```diff
@@ -39,6 +40,61 @@ import (
...
+		if exists && !isLedgerV3Preview(cluster) {
+			setLedgerV3Condition(ledger, metav1.ConditionFalse, "MigrationRequired", "A Ledger v3 Cluster exists; an explicit v3 to v2 migration is required")
+			return NewPendingError().WithMessage("migration required before switching Ledger from v3 to v2")
+		}
+		if previewVersion == "" {
+			if err := deleteLedgerV3Preview(ctx, stack); err != nil {
+				return err
+			}
+		}
+	}
```
NumaryBot, line 65: 🟠 **[major] Discovery failure treated as confirmed CRD absence, bypassing downgrade guard**

`ledgerV3ClusterAvailable == false` also represents discovery/list/watch failures. When this occurs, the code skips the existence check and proceeds with v2 reconciliation, potentially running legacy workloads alongside an inaccessible existing v3 Cluster instead of preserving the migration guard.

**Suggestion:** Distinguish confirmed CRD absence from discovery or access failures. When cluster access is unknown, preserve the migration guard and halt v2 reconciliation rather than assuming no Cluster exists.

internal/resources/ledgers/v3_preview.go
```diff
@@ -0,0 +1,175 @@
...
+	certificate := newLedgerV3Resource(ledgerV3CertificateGVK)
+	err = ctx.GetClient().Get(ctx, types.NamespacedName{Namespace: stack.Name, Name: ledgerV3TLSName(stack.Name)}, certificate)
+	if err == nil && certificate.GetLabels()[ledgerV3PreviewLabel] == "true" {
+		if err := ctx.GetClient().Delete(ctx, certificate); err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
+		secret := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{Namespace: stack.Name, Name: ledgerV3TLSName(stack.Name)}}
+		if err := ctx.GetClient().Delete(ctx, secret); err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
```
NumaryBot: 🟠 **[major] Preview TLS Secret not cleaned up when cert-manager is unavailable**

The early return when cert-manager is unavailable leaves the core TLS Secret behind, retaining preview private-key material. The later deletion is also conditional on the Certificate still existing, so the Secret can persist even after preview removal.

**Suggestion:** Look up and delete the Secret before the cert-manager capability check, guarded by its preview label or ownership check, ensuring private-key material is cleaned up regardless of cert-manager availability.

docs/04-Modules/03-Ledger.md
```diff
@@ -2,10 +2,15 @@ Formance Ledger is a real-time money tracking microservice that lets you model a
...
 
-Formance Ledger requires:
+Ledger versions up to and including `v3.0.0-alpha` require:
+
 - **PostgreSQL**: See configuration guide [here](../05-Infrastructure%20services/01-PostgreSQL.md).
 - (Optional) **Broker**: See configuration guide [here](../05-Infrastructure%20services/02-Message%20broker.md).
 
+Ledger versions newer than `v3.0.0-alpha` require the Ledger Operator and its
+`ledger.formance.com/v1alpha1` CRDs to be installed in the cluster. They use the
+Ledger v3 native storage and do not require a PostgreSQL `Database` resource.
```
NumaryBot: 🟠 **[major] cert-manager not documented as required dependency for Ledger v3**

`createOrUpdateV3TLSResources` keeps v3 pending unless both cert-manager `Issuer` and `Certificate` CRDs are discovered at startup. The requirements section currently mentions only the Ledger Operator, leading to incomplete deployments.

**Suggestion:** Update the requirements section to explicitly list cert-manager's `Issuer` and `Certificate` CRDs as prerequisites alongside the Ledger Operator CRDs.

internal/resources/ledgers/v3_spec.go
```diff
@@ -0,0 +1,221 @@
...
+	spec.Auth.Issuer = configuration.Issuer
+	if len(configuration.Issuers) > 0 {
+		spec.Auth.Issuers = slices.Clone(configuration.Issuers)
+	}
+	spec.Auth.CheckScopes = pointerTo(configuration.CheckScopes)
+	if configuration.CheckScopes {
+		spec.Auth.Service = configuration.Service
+	}
+	if configuration.ReadKeySetMaxRetries != 0 {
+		spec.Auth.ReadKeySetMaxRetries = pointerTo(int32(configuration.ReadKeySetMaxRetries))
```
NumaryBot, line 150: 🟡 **[minor] Missing bounds check before narrowing int to int32 for ReadKeySetMaxRetries**

`auths.ProtectedAuthConfiguration` stores `ReadKeySetMaxRetries` as an `int`. A direct cast to `int32` can silently overflow for values above `math.MaxInt32`, producing a negative retry count.

**Suggestion:** Add a bounds check before the cast and return an error or clamp the value if it exceeds `math.MaxInt32`.

deployment/operator/.gitignore
```diff
@@ -0,0 +1,2 @@
+deployment/operator/bin/
```
NumaryBot: 🟡 **[minor] bin/ ignore pattern is not relative to the containing directory**

The pattern `deployment/operator/bin/` in `.gitignore` is written as an absolute path from the repo root but `.gitignore` patterns are relative to their containing directory, so `deployment/operator/bin/` is not actually ignored.

**Suggestion:** Change the pattern to `bin/` so it correctly ignores the `bin/` directory relative to `deployment/operator/`.

docs/04-Modules/03-Ledger.md
```diff
@@ -2,10 +2,15 @@ Formance Ledger is a real-time money tracking microservice that lets you model a
 
 ## Requirements
 
-Formance Ledger requires:
+Ledger versions up to and including `v3.0.0-alpha` require:
+
 - **PostgreSQL**: See configuration guide [here](../05-Infrastructure%20services/01-PostgreSQL.md).
 - (Optional) **Broker**: See configuration guide [here](../05-Infrastructure%20services/02-Message%20broker.md).
```
NumaryBot: ⚪ **[nit] Non-descriptive 'here' link text**

Links using 'here' as link text are ambiguous out of context and fail accessibility and linting checks.

**Suggestion:** Replace 'here' with descriptive text such as 'PostgreSQL configuration' and 'message broker configuration'.

NumaryBot (commented):
NumaryBot posted 13 new inline findings.

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

internal/resources/ledgers/v3_preview.go
NumaryBot: 🟠 **[major] Gateway backend ports hard-coded instead of derived from LedgerConfiguration**

When a LedgerConfiguration overrides the Ledger Cluster service ports (e.g. spec.cluster.service.grpcPort or httpPort), composeLedgerV3ClusterSpec preserves those configured values, but the Gateway backend refs are still hard-coded to 9000/8888. This silently breaks HTTP/gRPC routing for any non-default port configuration.

**Suggestion:** Either enforce fixed default ports in the Cluster spec so they always match the hard-coded Gateway refs, or derive the Gateway backend port refs from the Cluster spec's desired port values so the two stay in sync.

internal/resources/ledgers/v3.go
NumaryBot, line 221: 🟠 **[major] Stale gRPC route left in place when TLS becomes unready (non-preview path)**

For the direct (non-preview) Ledger v3 path, if a GatewayGRPCAPI was previously created and the Certificate/Secret later becomes unready or is deleted, the reconciler returns pending without cleaning up the existing gRPC route. The Gateway can then continue exposing a backend whose TLS material is no longer valid. The preview path correctly deletes the route in this same TLS-pending scenario.

**Suggestion:** Mirror the preview path's behaviour: when the TLS precondition is not met (certificate/secret unready or missing), explicitly delete any existing GatewayGRPCAPI resource before returning pending.

internal/resources/ledgers/v3_preview.go
NumaryBot: 🟠 **[major] Preview TLS Secret not deleted independently of cert-manager availability**

When cert-manager is unavailable, the early return leaves the core TLS Secret behind. Later deletion is also conditional on the Certificate still existing. This retains preview private-key material after preview removal, leaking secrets.

**Suggestion:** Look up and delete the Secret before the capability return, guarded by its preview label or ownership check. Update the corresponding test (TestDeleteLedgerV3PreviewSkipsCertManagerResourcesWhenUnavailable) accordingly.

internal/resources/ledgers/init.go
```diff
@@ -39,6 +40,55 @@ import (
...
```
NumaryBot, line 65: 🟠 **[major] Downgrade guard bypassed when Cluster CRD access fails**

ledgerV3ClusterAvailable == false also covers discovery/list/watch failures, not just confirmed CRD absence. The current code skips the existence check and proceeds with v2 reconciliation in both cases, potentially running legacy v2 workloads alongside an inaccessible existing v3 Cluster.

**Suggestion:** Distinguish confirmed CRD absence from access/discovery failures. When Cluster access is unknown (i.e. an error occurred), preserve the migration guard and halt v2 reconciliation rather than assuming no Cluster exists.

internal/resources/ledgers/v3.go
```diff
@@ -0,0 +1,491 @@
...
+			certificateAvailable := watchLedgerV3Resource(ctx, b, options, crds, ledgerV3CertificateGVK)
+			ledgerV3CertManagerAvailable = issuerAvailable && certificateAvailable
+			return nil
+		})
+	}
+}
+
+func withLedgerConfigurationWatch() core.ReconcilerOption[*v1beta1.Ledger] {
+	return core.WithWatch[*v1beta1.Ledger, *v1beta1.LedgerConfiguration](
+		func(ctx core.Context, configuration *v1beta1.LedgerConfiguration) []reconcile.Request {
```
NumaryBot, line 100: 🟠 **[major] LedgerConfiguration watch handler does not reconcile previously matched ledgers on retarget**

When spec.stacks changes (e.g. from '*' to a specific stack, or from one stack to another), the watch handler only enqueues reconciles for the new targets. Ledgers that were matched by the previous target are not reconciled, so their v3 Cluster continues using the old shared configuration until an unrelated event triggers a reconcile.

**Suggestion:** Update the handler to account for both old and new targets on update events, or force a broader reconcile sweep when the stacks selector changes.

deployment/operator/main.go
NumaryBot: 🟠 **[major] licence-token retrieved as plaintext instead of secret**

cfg.Get("licence-token") downgrades the secret config value to plaintext before it is passed into Helm values and Pulumi state, exposing the licence token in state files and logs.

**Suggestion:** Use config.GetSecret(ctx, "licence-token") (or equivalent) and wrap it in pulumi.ToSecret so the value remains encrypted throughout the Pulumi graph.

deployment/operator/main.go
NumaryBot, line 45: 🟠 **[major] Silent no-op when no CRD manifests are found**

filepath.Glob returns an empty slice without error. Because Helm values disable chart-managed CRDs, an incorrect working directory or incomplete package silently deploys the operator without any CRDs installed.

**Suggestion:** Add a guard: if len(crdFiles) == 0 { return fmt.Errorf("no CRD files found under config/crd/bases") }

deployment/operator/helpers.go
NumaryBot, line 106: 🟠 **[major] Default arch filter silently drops arm64 from multi-arch images**

When arch is unset, it defaults to "amd64", filtering out linux-arm64. The resulting image cannot run on arm64 nodes, defeating the purpose of a multi-arch build.

**Suggestion:** Only apply the arch filter when arch is explicitly set; when unset, include all platforms.

deployment/operator/helpers.go
NumaryBot, line 165: 🟠 **[major] Cache mode appended to image Ref instead of typed field**

",mode=max" appended to the cache image reference makes the Ref invalid. The mode should be set via the typed CacheToRegistryArgs.Mode field.

**Suggestion:** Remove ",mode=max" from Ref and set Mode: "max" on CacheToRegistryArgs instead.

Dockerfile
NumaryBot, line 35: 🟠 **[major] Operator container runs as root**

The final image defaults to root, weakening containment if the operator process is compromised.

**Suggestion:** Add a non-root user (e.g. RUN addgroup -S operator && adduser -S -G operator operator) and switch to it with USER operator after copying the binary.

internal/resources/gateways/Caddyfile.gotpl
NumaryBot, line 185: 🟠 **[major] Health-check probes use http:// even when backend TLS is configured**

The /versions health-check probe URLs remain http:// even when HealthCheckBackend.TLS is configured. TLS-only backends will receive plaintext probes and appear unavailable.

**Suggestion:** Generate TLS-aware probe URLs (https:// when HealthCheckBackend.TLS is set) and apply the backend's configured CA trust settings to these probes.

docs/04-Modules/03-Ledger.md
NumaryBot: 🟠 **[major] cert-manager not documented as a required Ledger v3 dependency**

createOrUpdateV3TLSResources keeps v3 pending unless both cert-manager Issuer and Certificate CRDs were discovered at startup. The requirements section only mentions the Ledger Operator, leading operators to deploy incomplete stacks.

**Suggestion:** Add cert-manager (Issuer and Certificate CRDs) to the prerequisites section alongside the Ledger Operator.

internal/resources/ledgers/v3_spec.go
NumaryBot, line 150: 🟡 **[minor] Missing bounds check before int → int32 cast for ReadKeySetMaxRetries**

auths.ProtectedAuthConfiguration stores ReadKeySetMaxRetries as int. Casting it directly to int32 can silently overflow for values above math.MaxInt32, yielding a negative retry count.

**Suggestion:** Add a bounds check before the cast and return an error or clamp the value if it exceeds math.MaxInt32.

deployment/operator/.gitignore
NumaryBot: 🟡 **[minor] bin/ ignore pattern is not relative to the containing directory**

The pattern deployment/operator/bin/ in this .gitignore is treated as absolute from the repo root, not relative to deployment/operator/, so it may not ignore the intended directory.

**Suggestion:** Change the pattern to bin/ so it is correctly relative to deployment/operator/.

docs/04-Modules/03-Ledger.md
NumaryBot: 🟡 **[minor] Ambiguous 'here' link text in documentation**

Links with text 'here' are not understandable out of context and fail accessibility/markdown lint checks.

**Suggestion:** Replace 'here' with descriptive labels such as 'PostgreSQL configuration' and 'message broker configuration'.

NumaryBot (commented):
NumaryBot posted 15 new inline findings.

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

internal/resources/ledgers/init.go
flemzord: This behavior is intentional. Ledger v3 support is an optional capability: when discovery or access to the Ledger Operator CRD fails, the Formance Operator must keep reconciling the legacy Ledger v2 resources instead of turning a transient/missing optional dependency into an outage. The downgrade guard is therefore only enforced once the Ledger v3 capability has been positively established.

internal/resources/ledgers/v3.go
flemzord: No change is needed here. controller-runtime's  applies the mapper to both  and  for update events, so ledgers selected by the previous configuration and ledgers selected by the new configuration are both enqueued when  is retargeted.

internal/resources/ledgers/v3.go
```diff
@@ -0,0 +1,497 @@
...
+				return core.BuildReconcileRequests(
+					ctx,
+					ctx.GetClient(),
+					ctx.GetScheme(),
+					&v1beta1.Ledger{},
+				)
+			}
+
+			requests := make([]reconcile.Request, 0)
+			for _, stack := range configuration.GetStacks() {
```
NumaryBot, line 111: 🟠 **[major] LedgerConfiguration watch does not reconcile previously matched ledgers when stacks selector changes**

When a LedgerConfiguration update changes `spec.stacks`, the watch handler only enqueues reconciles for ledgers matched by the new selector. Ledgers matched by the old selector are never reconciled, so their v3 Cluster retains the stale removed configuration indefinitely until an unrelated event triggers a reconcile.

**Suggestion:** In the watch handler, also fetch and enqueue ledgers matched by the old `spec.stacks` selector (available from the Old object in the update event) so all previously-affected ledgers are reconciled when the selector changes.

internal/resources/ledgers/init.go
```diff
@@ -39,6 +40,48 @@ import (
...
```
NumaryBot, line 65: 🟠 **[major] Discovery failure treated as confirmed CRD absence, bypassing the v2/v3 downgrade guard**

`ledgerV3ClusterAvailable == false` covers both confirmed CRD absence and discovery/list/watch failures. The code at line 65 skips the v3 existence check and proceeds with v2 reconciliation in both cases, potentially running legacy v2 workloads alongside an inaccessible existing v3 Cluster and bypassing the migration guard.

**Suggestion:** Distinguish confirmed CRD absence (e.g. a typed 'resource not found' API-discovery error) from transient access/watch failures. In the failure case, preserve the migration guard and halt v2 reconciliation instead of assuming no Cluster exists.

internal/resources/ledgers/v3_preview.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func deleteLedgerV3Preview(ctx core.Context, stack *v1beta1.Stack) error {
+	cluster, exists, err := getV3Cluster(ctx, stack)
+	if err != nil {
+		return err
+	}
+	if exists && isLedgerV3Preview(cluster) {
+		if err := client.IgnoreNotFound(ctx.GetClient().Delete(ctx, cluster)); err != nil {
+			return err
```
NumaryBot, line 166: 🟠 **[major] Preview TLS Secret not cleaned up when cert-manager is unavailable**

The early-return on `ledgerV3CertManagerAvailable == false` leaves the core TLS Secret in place, retaining private-key material after preview removal. The later conditional deletion also depends on the Certificate still existing, meaning the Secret can persist even after preview removal.

**Suggestion:** Look up and delete the TLS Secret before the cert-manager capability guard, restricted by the Secret's preview label or ownership annotation, so private-key material is cleaned up regardless of cert-manager availability. Update the corresponding test accordingly.

internal/resources/gateways/Caddyfile.gotpl
```diff
@@ -133,16 +172,49 @@
...
+					{{- else }}
+					http://{{ $service.HealthCheckBackend.Name }}:{{ $service.HealthCheckBackend.Port }}/_info http://{{ $service.HealthCheckBackend.Name }}:{{ $service.HealthCheckBackend.Port }}/{{ $healthCheckEndpoint }}
+					{{- end }}
 				}
 				{{- else }}
-				{{ $service.Name }} http://{{ $service.Name }}:8080/_info http://{{ $service.Name }}:8080/{{ $service.HealthCheckEndpoint }}
+				{{- if $service.HealthCheckBackend.TLS }}
+				{{ $service.Name }} http://127.0.0.1:8080/_gateway-health/{{ $service.Name }}/_info http://127.0.0.1:8080/_gateway-health/{{ $service.Name }}/{{ $healthCheckEndpoint }}
+				{{- else }}
+				{{ $service.Name }} http://{{ $service.HealthCheckBackend.Name }}:{{ $service.HealthCheckBackend.Port }}/_info http://{{ $service.HealthCheckBackend.Name }}:{{ $service.HealthCheckBackend.Port }}/{{ $healthCheckEndpoint }}
```
NumaryBot, line 185: 🟠 **[major] Health-check probe URLs use http:// even for TLS-enabled backends**

The `/versions` health-check probes remain hardcoded to `http://` even when `HealthCheckBackend.TLS` is configured. TLS-only backends receive plaintext probes and may appear unavailable, degrading gateway routing.

**Suggestion:** Emit `https://` when `HealthCheckBackend.TLS` is set, and apply the backend's configured CA trust settings to these probes. Keep `http://` for non-TLS backends.

internal/resources/stacks/networkpolicies.go
```diff
@@ -96,11 +98,102 @@ func createNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
...
+			selector := directLedgerV3Selector(stack.Name)
+			np.Spec = networkingv1.NetworkPolicySpec{
+				PodSelector: selector,
+				PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
+				Ingress: []networkingv1.NetworkPolicyIngressRule{
+					{
+						From: []networkingv1.NetworkPolicyPeer{
+							{PodSelector: &selector},
+						},
+					},
```
NumaryBot, line 119: 🟠 **[major] Ledger v3 NetworkPolicy ports not derived from LedgerConfiguration overrides**

When a LedgerConfiguration overrides `spec.cluster.service.raftPort` or `grpcPort`, `composeLedgerV3ClusterSpec` uses the overridden ports but the NetworkPolicy still allows only the hard-coded defaults (7777 and 8888). Under NetworkPolicies, Raft/gRPC traffic between replicas is then denied, preventing cluster formation.

**Suggestion:** Derive the NetworkPolicy ingress ports from the same Cluster spec values used in `composeLedgerV3ClusterSpec` so the allowed ports always match the actual service ports.

deployment/operator/main.go
```diff
@@ -0,0 +1,126 @@
...
+		// Licence configuration
+		licenceIssuer := cfg.Get("licence-issuer")
+		if licenceIssuer == "" {
+			licenceIssuer = "https://license.formance.cloud/keys"
+		}
+		licenceToken := config.GetSecret(ctx, "licence-token")
+		licenceValues := pulumi.Map{
+			"createSecret": licenceToken.ApplyT(func(token string) bool {
+				return token != ""
+			}).(pulumi.BoolOutput),
```
NumaryBot, line 76: 🟠 **[major] Licence token retrieved as plaintext instead of a Pulumi secret**

`cfg.Get("licence-token")` retrieves the value as plaintext before passing it to `pulumi.String(...)`, stripping Pulumi's secret tracking. The token can then appear unencrypted in Pulumi state snapshots and logs.

**Suggestion:** Use `config.GetSecret(ctx, "licence-token")` to keep the value encrypted throughout the Pulumi state graph.

NumaryBot (commented):
NumaryBot posted 14 new inline findings.

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

deployment/operator/main.go
```diff
@@ -0,0 +1,126 @@
...
+		// Apply CRDs
+		crdFiles, err := filepath.Glob(filepath.Join("..", "..", "config", "crd", "bases", "*.yaml"))
+		if err != nil {
+			return fmt.Errorf("failed to glob CRD files: %w", err)
+		}
+		if len(crdFiles) == 0 {
+			return fmt.Errorf("no CRD manifests found under ../../config/crd/bases")
+		}
+
+		var crds []pulumi.Resource
```
NumaryBot, line 45: 🟠 **[major] CRD glob silently succeeds with zero files when working directory is wrong**

`filepath.Glob` returns an empty slice without error when no files match. Because Helm values disable chart-managed CRDs, an incorrect working directory or incomplete package would silently deploy the operator without any CRDs installed.

**Suggestion:** After the glob call, add: `if len(crdFiles) == 0 { return fmt.Errorf("no CRD files found under config/crd/bases") }`.

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    CGO_ENABLED=0 go build -buildvcs=false -o /usr/bin/operator -ldflags="${LDFLAGS}" .
+
+FROM alpine:3.20
+
+RUN apk update && apk add --no-cache ca-certificates curl
+RUN addgroup -S operator && adduser -S -G operator operator
+
+ENTRYPOINT ["/usr/bin/operator"]
+
+COPY --from=builder /usr/bin/operator /usr/bin/operator
```
NumaryBot, line 35: 🟠 **[major] Operator container runs as root**

The final image stage does not declare an unprivileged user, so the operator process runs as root. This weakens containment if the operator is ever compromised.

**Suggestion:** Add `RUN addgroup -S operator && adduser -S -G operator operator` after the `apk` install step and `USER operator` after copying the binary.

deployment/operator/helpers.go
```diff
@@ -0,0 +1,273 @@
...
+		platforms = platforms[:0]
+		for _, p := range allPlatforms {
+			if strings.HasSuffix(p, arch) {
+				platforms = append(platforms, p)
+			}
+		}
+		if len(platforms) == 0 {
+			platforms = []string{"linux-" + arch}
+		}
+	}
```
NumaryBot, line 106: 🟠 **[major] Default arch filter silently excludes arm64 from multi-arch images**

Defaulting `arch` to `amd64` when it is unset filters out the `linux/arm64` platform entry, making the 'multi-arch' image effectively amd64-only and unable to run on arm64 nodes.

**Suggestion:** Leave the platform list unfiltered when `arch` is not explicitly set; apply the suffix filter only when `arch` is explicitly provided by the caller.

deployment/operator/helpers.go
```diff
@@ -0,0 +1,273 @@
...
+					},
+				},
+			},
+			CacheTo: dockerbuild.CacheToArray{
+				dockerbuild.CacheToArgs{
+					Registry: dockerbuild.CacheToRegistryArgs{
+						Ref:  pulumi.Sprintf("%s/%s:buildcache-%s", dc.Registry, name, platform),
+						Mode: dockerbuild.CacheModeMax,
+					},
+				},
```
NumaryBot, line 165: 🟠 **[major] Cache mode appended to image Ref string instead of the typed field**

Appending `,mode=max` directly to the cache image `Ref` string produces an invalid image reference that most registries will reject. The cache mode must be set via the typed struct field.

**Suggestion:** Remove `,mode=max` from the `Ref` string and set `Mode: "max"` on the `CacheToRegistryArgs` struct instead.

Earthfile
```diff
@@ -45,9 +45,13 @@ compile:
...
 
 build-image:
-    FROM core+final-image
-    ENTRYPOINT ["/usr/bin/operator"]
-    COPY --pass-args (+compile/main) /usr/bin/operator
+    ARG LICENCE_PUBLIC_KEY_B64=""
+    ARG EARTHLY_BUILD_SHA
+    FROM DOCKERFILE \
+        --build-arg LICENCE_PUBLIC_KEY_B64=$LICENCE_PUBLIC_KEY_B64 \
+        --build-arg VERSION=$tag \
```
NumaryBot, line 52: 🟠 **[major] ARG tag used before it is declared in the build-image target**

In `+build-image`, `$tag` is referenced in Docker build args before the `ARG tag` declaration appears in the target. Depending on Earthly parsing order, the `VERSION` build arg can be empty or the target can fail when `--tag` is passed from `+deploy`.

**Suggestion:** Move `ARG tag=latest` before the `FROM DOCKERFILE` statement in the `+build-image` target.

docs/04-Modules/03-Ledger.md
```diff
@@ -2,9 +2,15 @@ Formance Ledger is a real-time money tracking microservice that lets you model a
...
-- **PostgreSQL**: See configuration guide [here](../05-Infrastructure%20services/01-PostgreSQL.md).
-- (Optional) **Broker**: See configuration guide [here](../05-Infrastructure%20services/02-Message%20broker.md).
+Ledger versions up to and including `v3.0.0-alpha` require:
+
+- See the [PostgreSQL configuration](../05-Infrastructure%20services/01-PostgreSQL.md).
+- (Optional) See the [message broker configuration](../05-Infrastructure%20services/02-Message%20broker.md).
+
+Ledger versions newer than `v3.0.0-alpha` require the Ledger Operator and its
+`ledger.formance.com/v1alpha1` CRDs, plus cert-manager with its `Issuer` and
+`Certificate` CRDs, to be installed in the cluster. They use Ledger v3 native
```
NumaryBot, line 12: 🟠 **[major] cert-manager not documented as a required Ledger v3 dependency**

The docs say Ledger v3 only requires the Ledger Operator and its CRDs. In practice, `createOrUpdateV3TLSResources` keeps v3 pending until cert-manager `Issuer` and `Certificate` CRDs are discovered, leading to incomplete deployments when cert-manager is absent.

**Suggestion:** Update the requirements section to mention cert-manager's `Issuer` and `Certificate` CRDs as required prerequisites alongside the Ledger Operator.

internal/resources/ledgers/v3_spec.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	spec.Auth.CheckScopes = pointerTo(configuration.CheckScopes)
+	if configuration.CheckScopes {
+		spec.Auth.Service = configuration.Service
+	}
+	if configuration.ReadKeySetMaxRetries != 0 {
+		if configuration.ReadKeySetMaxRetries > math.MaxInt32 || configuration.ReadKeySetMaxRetries < math.MinInt32 {
+			return fmt.Errorf("auth read key set max retries must fit in int32, got %d", configuration.ReadKeySetMaxRetries)
+		}
+		spec.Auth.ReadKeySetMaxRetries = pointerTo(int32(configuration.ReadKeySetMaxRetries))
```
NumaryBot, line 150: 🟡 **[minor] Unchecked narrowing cast of ReadKeySetMaxRetries to int32**

`auths.ProtectedAuthConfiguration` stores `ReadKeySetMaxRetries` as an `int`. Casting it directly to `int32` without a bounds check can silently overflow for large values, yielding a negative retry count.

**Suggestion:** Add a bounds check before the cast (e.g., verify the value is within `[0, math.MaxInt32]`) and return an error or clamp the value if it is out of range.

internal/tests/gateway_controller_test.go
```diff
@@ -289,6 +289,138 @@ var _ = Describe("GatewayController", func() {
...
+
+				Expect(LoadResource(stack.Name, tlsSecret.Name, tlsSecret)).To(Succeed())
+				patch := client.MergeFrom(tlsSecret.DeepCopy())
+				tlsSecret.Data["ca.crt"] = []byte("renewed CA")
+				Expect(Patch(tlsSecret, patch)).To(Succeed())
+
+				Eventually(func(g Gomega) string {
+					g.Expect(LoadResource(stack.Name, "gateway", deployment)).To(Succeed())
+					return deployment.Spec.Template.Annotations["formance.com/backend-tls-secrets-hash"]
+				}).ShouldNot(Equal(initialHash))
```
NumaryBot: 🟡 **[minor] Rollout hash assertion passes when annotation is removed**

`ShouldNot(Equal(initialHash))` also passes when the annotation is removed and returns `""`, which would mask a broken rollout hash implementation.

**Suggestion:** Replace with `Should(SatisfyAll(Not(BeEmpty()), Not(Equal(initialHash))))` to verify the hash is both non-empty and changed.

internal/resources/ledgers/v3.go
NumaryBot, line 111: 🟠 **[major] LedgerConfiguration watch does not reconcile previously matched ledgers when stacks selector changes**

When a LedgerConfiguration update changes spec.stacks, the watch handler only enqueues reconciles for ledgers matched by the new selector. Ledgers matched by the old selector are never reconciled, so their v3 Cluster retains the stale removed configuration indefinitely until an unrelated event triggers a reconcile. The maintainer dismissed this by claiming controller-runtime applies the mapper to both old and new objects, but the inline thread shows ongoing contention that the old targets are not properly swept.

**Suggestion:** In the watch handler, also fetch and enqueue ledgers matched by the old spec.stacks selector (available from the Old object in the update event) so all previously-affected ledgers are reconciled when the selector changes.

internal/resources/ledgers/v3_preview.go
NumaryBot, line 166: 🟠 **[major] Preview TLS Secret not cleaned up when cert-manager is unavailable**
<sub>reported by NumaryBot, codex</sub>

The early-return on ledgerV3CertManagerAvailable == false leaves the core TLS Secret in place, retaining private-key material after preview removal. The later conditional deletion also depends on the Certificate still existing, meaning the Secret can persist even after preview removal.

**Suggestion:** Look up and delete the TLS Secret before the cert-manager capability guard, restricted by the Secret's preview label or ownership annotation, so private-key material is cleaned up regardless of cert-manager availability. Update the corresponding test accordingly.

internal/resources/gateways/Caddyfile.gotpl
NumaryBot, line 185: 🟠 **[major] Health-check probe URLs use http:// even for TLS-enabled backends**
<sub>reported by NumaryBot, codex</sub>

The /versions health-check probes remain hardcoded to http:// even when HealthCheckBackend.TLS is configured. TLS-only backends receive plaintext probes and may appear unavailable, degrading gateway routing.

**Suggestion:** Emit https:// when HealthCheckBackend.TLS is set, and apply the backend's configured CA trust settings to these probes. Keep http:// for non-TLS backends.

internal/resources/ledgers/v3.go
```diff
@@ -0,0 +1,497 @@
...
+		setLedgerV3Condition(ledger, metav1.ConditionFalse, "TLSReconcileFailed", err.Error())
+		return err
+	}
+
+	cluster, clusterSpec, err := createOrUpdateV3Cluster(ctx, stack, ledger, version, false, tlsCAHash)
+	if err != nil {
+		setLedgerV3Condition(ledger, metav1.ConditionFalse, "ReconcileFailed", err.Error())
+		return err
+	}
+	if !tlsReady {
```
NumaryBot, line 221: 🟠 **[major] Stale gRPC route left in place when TLS becomes unready on the non-preview path**

For the direct (non-preview) Ledger v3 path, if a GatewayGRPCAPI was previously created and the Certificate/Secret later becomes unready or is deleted, the reconciler returns pending without cleaning up the existing gRPC route. The Gateway can then continue exposing a backend whose TLS material is no longer valid. The preview path correctly deletes the route in this same TLS-pending scenario.

**Suggestion:** Mirror the preview path's behaviour: when the TLS precondition is not met (certificate/secret unready or missing), explicitly delete any existing GatewayGRPCAPI resource before returning pending.

internal/resources/stacks/networkpolicies.go
NumaryBot, line 119: 🟠 **[major] Ledger v3 NetworkPolicy ports hard-coded instead of derived from LedgerConfiguration**

When a LedgerConfiguration overrides spec.cluster.service.raftPort or grpcPort, composeLedgerV3ClusterSpec uses the overridden ports but the NetworkPolicy still allows only the hard-coded defaults (7777 and 8888). Under NetworkPolicies, Raft/gRPC traffic between replicas is then denied, preventing cluster formation.

**Suggestion:** Derive the NetworkPolicy ingress ports from the same Cluster spec values used in composeLedgerV3ClusterSpec so the allowed ports always match the actual service ports.

internal/resources/ledgers/v3_preview.go
```diff
@@ -0,0 +1,201 @@
...
+	); err != nil {
+		setLedgerV3PreviewCondition(ledger, metav1.ConditionFalse, "GatewayReconcileFailed", err.Error())
+		return err
+	}
+
+	ready, message, err := isV3ClusterReady(cluster)
+	if err != nil {
+		return err
+	}
+	if !ready {
```
NumaryBot, line 109: 🟠 **[major] Gateway backend ports hard-coded instead of derived from LedgerConfiguration**

When a LedgerConfiguration overrides the Ledger Cluster service ports (e.g. spec.cluster.service.grpcPort or httpPort), composeLedgerV3ClusterSpec preserves those configured values, but the Gateway backend refs are still hard-coded to 9000/8888. This silently breaks HTTP/gRPC routing for any non-default port configuration.

**Suggestion:** Either enforce fixed default ports in the Cluster spec so they always match the hard-coded Gateway refs, or derive the Gateway backend port refs from the Cluster spec's desired port values so the two stay in sync.

deployment/operator/main.go
NumaryBot, line 76: 🟠 **[major] Licence token retrieved as plaintext instead of a Pulumi secret**
<sub>reported by NumaryBot, codex</sub>

cfg.Get("licence-token") retrieves the value as plaintext before passing it to pulumi.String(...), stripping Pulumi's secret tracking. The token can then appear unencrypted in Pulumi state snapshots and logs.

**Suggestion:** Use config.GetSecret(ctx, "licence-token") to keep the value encrypted throughout the Pulumi state graph.

deployment/operator/main.go
NumaryBot, line 45: 🟠 **[major] CRD glob silently succeeds with zero files when working directory is wrong**
<sub>reported by NumaryBot, codex</sub>

filepath.Glob returns an empty slice without error when no files match. Because Helm values disable chart-managed CRDs, an incorrect working directory or incomplete package would silently deploy the operator without any CRDs installed.

**Suggestion:** After the glob call, add: if len(crdFiles) == 0 { return fmt.Errorf("no CRD files found under config/crd/bases") }.

deployment/operator/helpers.go
NumaryBot, line 106: 🟠 **[major] Default arch filter silently excludes arm64 from multi-arch images**

Defaulting arch to "amd64" when it is unset filters out the linux/arm64 platform entry, making the 'multi-arch' image effectively amd64-only and unable to run on arm64 nodes.

**Suggestion:** Leave the platform list unfiltered when arch is not explicitly set; apply the suffix filter only when arch is explicitly provided by the caller.

deployment/operator/helpers.go
NumaryBot, line 165: 🟠 **[major] Cache mode appended to image Ref string instead of typed field**

",mode=max" appended directly to the cache image reference string produces an invalid image reference. The cache mode should be set via the typed CacheToRegistryArgs.Mode field.

**Suggestion:** Remove ",mode=max" from the Ref string and set Mode: "max" on the CacheToRegistryArgs struct instead.

Dockerfile
NumaryBot, line 35: 🟠 **[major] Operator container runs as root**
<sub>reported by NumaryBot, codex</sub>

The final image stage does not declare an unprivileged user, so the operator process runs as root. This weakens containment if the operator is ever compromised.

**Suggestion:** Add RUN addgroup -S operator && adduser -S -G operator operator after the apk install step and USER operator after copying the binary.

docs/04-Modules/03-Ledger.md
NumaryBot, line 12: 🟠 **[major] cert-manager not documented as a required Ledger v3 dependency**

createOrUpdateV3TLSResources keeps v3 pending unless both cert-manager Issuer and Certificate CRDs were discovered at startup. The requirements section only mentions the Ledger Operator, leading operators to deploy incomplete stacks.

**Suggestion:** Add cert-manager (Issuer and Certificate CRDs) to the prerequisites section alongside the Ledger Operator.

api/formance.com/v1beta1/ledgerconfiguration_types.go
```diff
@@ -0,0 +1,70 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec LedgerConfigurationSpec `json:"spec,omitempty"`
+}
+
+func (in *LedgerConfiguration) GetStacks() []string {
+	return in.Spec.Stacks
+}
+
+func (in *LedgerConfiguration) IsWildcard() bool {
```
NumaryBot, line 56: 🟠 **[major] Mixed wildcard and explicit stacks selector not handled correctly**

If a LedgerConfiguration uses spec.stacks: ["*", "some-stack"], IsWildcard() returns false because the slice contains more than just "*". The watch handler then only enqueues ledgers matching the explicit stack name, missing all other stacks that should also be reconciled due to the wildcard. Either the mixed form should be rejected at admission time or any selector containing "*" should be treated as global.

**Suggestion:** Either reject mixed selectors (containing both "*" and explicit stack names) via a webhook or validation rule, or update IsWildcard() and the watch handler so that any selector containing "*" triggers a global reconcile sweep.

internal/resources/ledgers/v3_spec.go
NumaryBot, line 150: 🟡 **[minor] Missing bounds check before int → int32 cast for ReadKeySetMaxRetries**

auths.ProtectedAuthConfiguration stores ReadKeySetMaxRetries as int. Casting it directly to int32 can silently overflow for values above math.MaxInt32, yielding a negative retry count.

**Suggestion:** Add a bounds check before the cast and return an error or clamp the value if it exceeds math.MaxInt32.

deployment/operator/.gitignore
```diff
@@ -0,0 +1,2 @@
+bin/
```
NumaryBot, line 1: 🟡 **[minor] bin/ ignore pattern is not relative to the containing directory**

The pattern deployment/operator/bin/ in this .gitignore is treated as an absolute path from the repo root, not relative to deployment/operator/, so it may not ignore the intended directory.

**Suggestion:** Change the pattern to bin/ so it is correctly relative to deployment/operator/.

docs/04-Modules/03-Ledger.md
```diff
@@ -2,9 +2,15 @@ Formance Ledger is a real-time money tracking microservice that lets you model a
 
 ## Requirements
 
-Formance Ledger requires:
-- **PostgreSQL**: See configuration guide [here](../05-Infrastructure%20services/01-PostgreSQL.md).
-- (Optional) **Broker**: See configuration guide [here](../05-Infrastructure%20services/02-Message%20broker.md).
+Ledger versions up to and including `v3.0.0-alpha` require:
+
+- See the [PostgreSQL configuration](../05-Infrastructure%20services/01-PostgreSQL.md).
+- (Optional) See the [message broker configuration](../05-Infrastructure%20services/02-Message%20broker.md).
```
NumaryBot, line 8: ⚪ **[nit] Non-descriptive 'here' link text in documentation**

Links using 'here' as link text are ambiguous out of context and fail accessibility and markdown lint checks.

**Suggestion:** Replace 'here' with descriptive labels such as 'PostgreSQL configuration' and 'message broker configuration'.

NumaryBot (commented):
NumaryBot posted 16 new inline findings.

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

internal/resources/gateways/init.go
```diff
@@ -160,7 +176,16 @@ func init() {
 			}),
 			WithWatchSettings[*v1beta1.Gateway](),
 			WithWatchDependency[*v1beta1.Gateway](&v1beta1.GatewayHTTPAPI{}),
+			WithWatchDependency[*v1beta1.Gateway](&v1beta1.GatewayGRPCAPI{}),
 			WithWatchDependency[*v1beta1.Gateway](&v1beta1.Auth{}),
+			WithWatch[*v1beta1.Gateway](func(ctx Context, secret *corev1.Secret) []reconcile.Request {
```
NumaryBot, line 181: 🟠 **[major] Watch does not cover unlabeled backend TLS Secrets referenced by backendRef.tls**

The Secret watch handler only reacts to Secrets carrying the internal `formance.com/gateway-backend-tls` label. However, the public `GatewayHTTPAPI`/`GatewayGRPCAPI` CRD allows users to reference any Secret via `backendRef.tls`. When a referenced CA or cert is rotated in such an unlabeled Secret, the Gateway pod continues using stale trust material until an unrelated reconcile event happens, which can cause TLS verification failures against the backend.

**Suggestion:** Either enforce or document the `formance.com/gateway-backend-tls` label requirement in the CRD/docs, or change the watch predicate to also index and react to Secrets that are directly referenced by active `GatewayHTTPAPI`/`GatewayGRPCAPI` backend refs (e.g. via a field indexer on those CRDs), so certificate rotations always trigger a Gateway reconcile.

internal/resources/ledgers/v3.go
```diff
@@ -0,0 +1,497 @@
...
+		Status:             status,
+		ObservedGeneration: ledger.GetGeneration(),
+		LastTransitionTime: metav1.Now(),
+		Reason:             reason,
+		Message:            message,
+	}, v1beta1.ConditionTypeMatch(ledgerV3ClusterReadyCondition))
+}
+
+func legacyLedgerResourcesExist(ctx core.Context, stack *v1beta1.Stack) (bool, error) {
+	for _, name := range []string{"ledger", "ledger-worker", "ledger-write", "ledger-read", "ledger-gateway"} {
```
NumaryBot, line 284: 🟠 **[major] v2 migration Jobs not included in the v2-resource-existence guard**

When v2 migration Jobs (e.g. `ledger-migration`, `ledger-create-database`) still exist after the Deployment/Database objects have been removed, the guard check returns false and `reconcileV3` proceeds to create the v3 Cluster. This bypasses the stated guard against switching generations while v2 resources still exist, risking v2 jobs mutating shared state alongside the new v3 cluster.

**Suggestion:** Extend the migration guard to also check for the existence of v2 batch/Job resources before allowing v3 Cluster creation. Only proceed when all v2 resource types—including Jobs—are confirmed absent.

internal/resources/ledgers/v3_tls.go
```diff
@@ -0,0 +1,204 @@
...
+	labels[v1beta1.StackLabel] = stackName
+	if preview {
+		labels[ledgerV3PreviewLabel] = "true"
+	} else {
+		delete(labels, ledgerV3PreviewLabel)
+	}
+	resource.SetLabels(labels)
+}
+
+func ledgerV3IssuerName(stackName string) string {
```
NumaryBot, line 120: 🟠 **[major] Issuer name not scoped to Ledger v3, risking collision with existing Issuers**

The Issuer is created with the generic name `<stack>-selfsigned`. In namespaces that already contain an unrelated Issuer with that same name, this path will either overwrite/adopt that Issuer or fail if it is already controller-owned by another resource. The TLS Secret was given a dedicated Ledger v3–specific name precisely to avoid such collisions; the Issuer should follow the same pattern.

**Suggestion:** Give the Issuer a Ledger v3–specific name (e.g. `<stack>-ledger-v3-selfsigned`) to avoid collisions with pre-existing Issuers in the same namespace.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

NumaryBot (commented):

Resolved 2 stale NumaryBot review threads (2 fixed, 0 outdated).

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

internal/resources/ledgers/v3.go
```diff
@@ -0,0 +1,513 @@
...
+		setLedgerV3Condition(ledger, metav1.ConditionFalse, "TLSReconcileFailed", err.Error())
+		return err
+	}
+
+	cluster, clusterSpec, err := createOrUpdateV3Cluster(ctx, stack, ledger, version, false, tlsCAHash)
+	if err != nil {
+		setLedgerV3Condition(ledger, metav1.ConditionFalse, "ReconcileFailed", err.Error())
+		return err
+	}
+	if !tlsReady {
```
NumaryBot, line 221: 🟠 **[major] Stale direct HTTP gateway route not cleaned up when TLS becomes unready**

On the non-preview path, if the v3 Ledger was previously reconciled successfully and the Certificate/Secret later becomes unready or is deleted, the reconciler removes only the GatewayGRPCAPI and returns before cleaning up the GatewayHTTPAPI. The existing HTTP route (`/api/ledger`) continues pointing at the v3 backend, exposing a cluster whose TLS precondition has failed. The preview path already handles this correctly by deleting the route before returning pending.

**Suggestion:** Mirror the preview path's behaviour: when the TLS precondition is not met (certificate/secret unready or missing), explicitly delete any existing GatewayHTTPAPI resource in addition to the GatewayGRPCAPI before returning pending.

internal/resources/ledgers/v3_preview.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		setLedgerV3PreviewCondition(ledger, metav1.ConditionFalse, "TLSReconcileFailed", err.Error())
+		return err
+	}
+	cluster, clusterSpec, err := createOrUpdateV3Cluster(ctx, stack, ledger, version, true, tlsCAHash)
+	if err != nil {
+		setLedgerV3PreviewCondition(ledger, metav1.ConditionFalse, "ReconcileFailed", err.Error())
+		return err
+	}
+	if !tlsReady {
```
NumaryBot, line 70: 🟠 **[major] Stale preview HTTP gateway route not cleaned up when TLS becomes unready**

After a preview route has been published once, if the v3 Certificate/Secret later becomes unready the reconciler deletes only the GatewayGRPCAPI and returns before updating or removing the GatewayHTTPAPI. The previous `/api/ledger/v3` rule is left in place, allowing the gateway to continue routing to the preview backend despite the TLS readiness gate failing.

**Suggestion:** Before the early return on TLS unreadiness, also delete (or gate) the existing GatewayHTTPAPI resource for the preview route, consistent with how the direct path handles the analogous GatewayGRPCAPI cleanup.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

internal/resources/gatewaygrpcapis/init.go
```diff
@@ -0,0 +1,58 @@
...
+
+	v1beta1 "github.com/formancehq/operator/v3/api/formance.com/v1beta1"
+	. "github.com/formancehq/operator/v3/internal/core"
+	"github.com/formancehq/operator/v3/internal/resources/services"
+)
+
+//+kubebuilder:rbac:groups=formance.com,resources=gatewaygrpcapis,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=formance.com,resources=gatewaygrpcapis/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=formance.com,resources=gatewaygrpcapis/finalizers,verbs=update
+
```
NumaryBot, line 30: 🟠 **[major] gRPC backend Service deleted even when explicitly referenced via spec.backendRef**

When spec.backendRef is set to the same Service name as the historical default (<name>-grpc), the reconcile path deletes the Service that the route is explicitly configured to use. This causes valid configurations — such as adding TLS or a non-default port via backendRef to an existing <name>-grpc Service — to break on every reconciliation. Cleanup should be restricted to Services owned by this GatewayGRPCAPI.

**Suggestion:** Limit Service deletion to resources that are owned by (or were created by) this GatewayGRPCAPI controller, rather than matching solely by conventional name. Check ownership annotations/labels before deleting.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

internal/resources/gatewaygrpcapis/init.go
flemzord: Out of scope per maintainer direction. The Ledger v3 integration does not hit this case: its explicit backendRef targets ledger-<stack>, not the historical <name>-grpc Service. The behavior of manually authored public GatewayGRPCAPI resources will be handled separately.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

api/formance.com/v1beta1/gatewayhttpapi_types.go
```diff
@@ -27,6 +27,9 @@ type GatewayHTTPAPIRule struct {
 	//+optional
 	//+kubebuilder:default:=false
 	Secured bool `json:"secured"`
+	// BackendRef overrides the historical module Service for this rule.
```
NumaryBot, line 30: 🟠 **[major] BackendRef with same name as spec.name is silently overwritten by the HTTP API reconciler**

When a GatewayHTTPAPI rule uses backendRef with a name equal to spec.name (e.g., to change the port or enable backend TLS while keeping the conventional Service name), the HTTP API reconciler still unconditionally creates/updates the Service named spec.name with the default 8080 configuration on every reconcile cycle. This breaks any valid override that reuses the conventional name, as the controller silently clobbers the user-provided configuration.

**Suggestion:** Either reject at admission time any backendRef whose name matches spec.name (to prevent ambiguity), or update the reconciler to skip managing the legacy default Service when all rules are fully backed by backendRef entries, so user-provided Service configurations are not overwritten.

internal/resources/gatewaygrpcapis/init.go
```diff
@@ -0,0 +1,58 @@
...
+func Reconcile(ctx Context, _ *v1beta1.Stack, grpcAPI *v1beta1.GatewayGRPCAPI) error {
+	if grpcAPI.Spec.BackendRef != nil {
```
NumaryBot, line 32: 🟠 **[major] Missing admission guard for backendRef name collision with managed gRPC Service**

When a GatewayGRPCAPI sets spec.backendRef.name to the conventional managed Service name (<spec.name>-grpc), the reconciler deletes that same Service and the generated Caddy route then points at a missing backend. The HTTP API added an equivalent admission guard for this collision, but the new gRPC API has no analogous validation. Users reusing the conventional Service name for a backendRef override can silently break gRPC routing on every reconcile cycle.

**Suggestion:** Add an admission validation (webhook or reconciler-level rejection) that rejects any GatewayGRPCAPI whose spec.backendRef.name matches the conventional managed Service name (<spec.name>-grpc), mirroring the guard already in place for GatewayHTTPAPI.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/490#issuecomment-4981366834

internal/resources/gatewaygrpcapis/init.go
flemzord: Out of scope per maintainer direction, consistently with the earlier public GatewayGRPCAPI finding. Ledger v3 uses ledger-<stack> as its backendRef and never reuses the historical <name>-grpc Service, so this integration is unaffected.

NumaryBot (approved):

PR #492, opened 2026-07-16
feat(connectivity): add Connectivity module bound to the stack ledger

Dav-14:
## What

Adds a new **`formance.com/v1beta1 Connectivity`** stack module that binds connectivity to the stack's ledger, mirroring the Ledger v3 delegation pattern.

The module doesn't run the workload itself — it delegates to the connectivity operator (`connectivity.formance.com`) and reflects its readiness, exactly like the `Ledger` module delegates to the ledger operator (`ledger.formance.com Cluster`).

## Behaviour

1. **Capability + API-group detection** (same mechanism as Ledger v3): at controller start-up it lists CRDs, checks the `connectivity.formance.com/Connectivity` CRD is present with a served version, and runs a `SelfSubjectAccessReview` for each required verb. If the group/CRD/RBAC is missing it reports the capability as unavailable and stays **pending** — it never fails controller setup.
2. **Gates on Ledger v3**: connectivity ingests into the Ledger v3 gRPC endpoint, so it only provisions once the stack's `Ledger` module is v3 and ready.
3. **Binds to the stack ledger**: provisions a `connectivity.formance.com/v1alpha1 Connectivity` with
   - `ledgerAddress` = the ledger v3 gRPC service, and
   - `ledgerTLS` = the ledger backend TLS secret (CA + SNI).

   The connection details come from `ledgers.V3GRPCBackendRef` — the single source of truth already used to reach the ledger over gRPC — so connectivity and the gateway stay in sync (no duplicated address/secret naming).
4. Reflects the delegated resource's `status.phase` back onto the module's Ready condition.

## Files

- `api/formance.com/v1beta1/connectivity_types.go` — new module CR (`Connectivity`, labelled `formance.com/kind=module`).
- `internal/resources/connectivities/init.go` — reconciler, capability detection, ledger-v3 gate, delegated-resource bind.
- `internal/resources/connectivities/init_test.go` — unit tests for the capability-gating paths (discovery failure, inaccessible resource, missing RBAC, capability-unavailable reconcile).
- `internal/resources/ledgers/exports.go` — exports `IsV3` + `V3GRPCBackendRef` so the ledger v3 gRPC connection stays the single source of truth.
- `internal/resources/all.go` — registers the module.
- Generated: CRD base, RBAC role, deepcopy, helm CRD + operator ClusterRole.

## Test

- `go build ./...` ✅
- `go vet ./internal/resources/connectivities/...` ✅
- `go test ./internal/resources/connectivities/...` ✅
- `helm template ./helm/crds` renders ✅

## Follow-ups (out of scope)

- Wiring the connectivity image / registry rewrite + imagePullSecrets from stack settings (currently left to the connectivity operator defaults).
- Membership/fctl surface to enable the module on a stack.


internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,252 @@
...
+	ledger, err := getStackLedger(ctx, stack.Name)
+	if err != nil {
+		return err
+	}
+	if ledger == nil {
+		setCondition(connectivity, metav1.ConditionFalse, "LedgerNotFound", "stack has no Ledger module")
+		return NewPendingError().WithMessage("connectivity requires a Ledger module on the stack")
+	}
+
+	ledgerVersion := ledger.Spec.Version
```
NumaryBot: 🟠 **[major] Resolve the ledger version through Versions resources**

For stacks that use `.spec.versionsFromFile` instead of `.spec.version`, this fallback never reads the Ledger entry from the referenced `Versions` object, so `ledgerVersion` stays empty and Connectivity remains stuck with `LedgerNotV3` even when the Ledger module itself reconciled as v3 and is ready. Use the same module-version resolution path as the Ledger reconciler for the Ledger object before applying the v3 gate.

config/crd/bases/formance.com_connectivities.yaml
```diff
@@ -0,0 +1,153 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.18.0
+  labels:
+    formance.com/kind: module
+  name: connectivities.formance.com
```
NumaryBot, line 9: 🟠 **[major] Add the Connectivity CRD to kustomize resources**

This new CRD is generated, but `config/crd/kustomization.yaml` does not reference `bases/formance.com_connectivities.yaml`, so deployments that install CRDs through the existing kustomize path will not create the `formance.com/v1beta1 Connectivity` API even though the controller and RBAC are present. Add this base to the kustomization resources list so non-Helm installs can create Connectivity modules.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

NumaryBot (commented):

Resolved 2 stale NumaryBot review threads (1 fixed, 1 outdated).

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,256 @@
...
+			log.FromContext(ctx).Info("Connectivity dependency permission is unavailable; continuing without it", "gvk", gvk, "verb", verb, "reason", review.Status.Reason)
+			return false
+		}
+	}
+	return true
+}
+
+func init() {
+	Init(
+		WithModuleReconciler(Reconcile,
```
NumaryBot: 🟡 **[minor] Connectivity version resolution fails on versionsFromFile stacks without a connectivity entry**

For stacks that use spec.versionsFromFile, the module framework resolves the Connectivity version before entering Reconcile. If the referenced Versions object has no 'connectivity' key, reconciliation fails before reaching the capability/Ledger-v3 gate, even though the Connectivity reconciler never actually uses the resolved version value. This forces operators to add a meaningless version entry to their Versions file just to unblock reconciliation.

**Suggestion:** Make the version field optional or provide a sensible default (e.g. empty string) so that the reconciler can proceed when no connectivity version is specified. If the version is genuinely unused, skip the version-resolution gate entirely for this module.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

internal/resources/connectivities/init.go
NumaryBot: 🟡 **[minor] Connectivity version resolution fails on versionsFromFile stacks without a 'connectivity' entry**

For stacks that use spec.versionsFromFile, the module framework resolves the Connectivity version before entering Reconcile. If the referenced Versions object has no 'connectivity' key, reconciliation fails before reaching the capability/Ledger-v3 gate, even though the resolved version value is never actually used by the reconciler. This forces operators to add a meaningless version entry to their Versions file just to unblock reconciliation.

**Suggestion:** Make the version field optional or provide a sensible default (e.g. empty string) so the reconciler can proceed when no connectivity version is specified. If the version is genuinely unused, skip the version-resolution gate entirely for this module.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,425 @@
...
+	// Provision a god-mode ledger credential so connectivity-core can
+	// authenticate its gRPC calls. The ledger operator registers the public key
+	// on the ledger and distributes the private seed as a Secret in the stack
+	// namespace; connectivity-core is wired to it via spec.auth below.
+	authKeyID, authSecretName, credReady, err := ensureLedgerCredentials(ctx, stack)
+	if err != nil {
+		setCondition(connectivity, metav1.ConditionFalse, "LedgerCredentialsFailed", err.Error())
+		return err
+	}
+	if !credReady {
```
NumaryBot, line 195: 🟠 **[major] No watch on ledger Credentials — reconciler can stall indefinitely**

When the Connectivity reconciler creates the ledger Credentials object and then returns a pending/application error waiting for `status.phase=Ready`, it will not be requeued automatically because the Credentials resource is not registered as a watched object. The reconciler only watches GatewayHTTPAPI, the delegated Connectivity, settings, and Ledger. On a normal install the controller can remain stuck in `LedgerCredentialsPending` until some unrelated watched object triggers a reconcile.

**Suggestion:** Add a Watch for the ledger Credentials resource in the controller setup so that a status update on that object (e.g. phase→Ready) triggers a reconcile of the owning Connectivity. Alternatively, use a periodic requeue (e.g. `ctrl.Result{RequeueAfter: ...}`) whenever the reconciler returns while waiting for Credentials readiness.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,426 @@
...
```
NumaryBot, line 195: 🟠 **[major] No watch on ledger Credentials — reconciler can stall indefinitely**

When the Connectivity reconciler creates the ledger Credentials object and then returns a pending/application error waiting for `status.phase=Ready`, it will not be requeued automatically because the Credentials resource is not registered as a watched object. The reconciler only watches GatewayHTTPAPI, the delegated Connectivity, settings, and Ledger. On a normal install the controller can remain stuck in `LedgerCredentialsPending` until some unrelated watched object triggers a reconcile.

**Suggestion:** Add a Watch for the ledger Credentials resource in the controller setup so that a status update on that object (e.g. phase→Ready) triggers a reconcile of the owning Connectivity. Alternatively, use a periodic requeue (e.g. `ctrl.Result{RequeueAfter: someInterval}`) whenever the reconciler returns while waiting for Credentials readiness.

internal/resources/stacks/networkpolicies.go
```diff
@@ -96,11 +98,104 @@ func createNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
...
+			Name:      "allow-ledger-v3-cluster",
+		},
+		func(np *networkingv1.NetworkPolicy) error {
+			selector := directLedgerV3Selector(stack.Name)
+			np.Spec = networkingv1.NetworkPolicySpec{
+				PodSelector: selector,
+				PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
+				Ingress: []networkingv1.NetworkPolicyIngressRule{
+					{
+						From: []networkingv1.NetworkPolicyPeer{
```
NumaryBot, line 137: 🟠 **[major] Connectivity pods blocked by Ledger v3 NetworkPolicy when network policies are enabled**

When `networkpolicies.enabled` is true and Connectivity is enabled, the delegated connectivity workload dials the Ledger directly at `ledger-<stack>:8888` (gRPC). The new Ledger-v3 ingress NetworkPolicy only allows port 8888 from pods matching the Ledger selector, so Connectivity pods — which are neither gateway nor Ledger pods — have their gRPC connections silently dropped. This would manifest as a hard runtime failure on any stack with network policies enabled.

**Suggestion:** Add an ingress rule to the Ledger v3 NetworkPolicy (or a dedicated NetworkPolicy) that allows pods matching the Connectivity workload pod selector to reach the Ledger gRPC port (8888).

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

internal/resources/connectivities/init.go
NumaryBot, line 195: 🟠 **[major] No watch on ledger Credentials — reconciler can stall indefinitely**

When the Connectivity reconciler creates the ledger Credentials object and then returns a pending/application error waiting for status.phase=Ready, it will not be requeued automatically because the Credentials resource is not registered as a watched object. The reconciler only watches GatewayHTTPAPI, the delegated Connectivity, settings, and Ledger. On a normal install the controller can remain stuck in LedgerCredentialsPending until some unrelated watched object triggers a reconcile.

**Suggestion:** Add a Watch for the ledger Credentials resource in the controller setup so that a status update on that object (e.g. phase→Ready) triggers a reconcile of the owning Connectivity. Alternatively, use a periodic requeue (e.g. ctrl.Result{RequeueAfter: someInterval}) whenever the reconciler returns while waiting for Credentials readiness.

internal/resources/stacks/networkpolicies.go
NumaryBot, line 137: 🟠 **[major] Connectivity pods blocked by Ledger v3 NetworkPolicy when network policies are enabled**

When networkpolicies.enabled is true and Connectivity is enabled, the delegated connectivity workload dials the Ledger directly at ledger-<stack>:8888 (gRPC). The new Ledger-v3 ingress NetworkPolicy only allows port 8888 from pods matching the Ledger selector, so Connectivity pods — which are neither gateway nor Ledger pods — have their gRPC connections silently dropped. This manifests as a hard runtime failure on any stack with network policies enabled.

**Suggestion:** Add an ingress rule to the Ledger v3 NetworkPolicy (or a dedicated NetworkPolicy) that allows pods matching the Connectivity workload pod selector to reach the Ledger gRPC port (8888).

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,426 @@
...
+// Stack rather than the namespaced Connectivity — a namespaced owner cannot own
+// a cluster-scoped resource. The ledger operator generates the Ed25519 keypair,
+// registers the public key on the Cluster matched by the stack selector, and
+// distributes the private seed as a Secret into the stack namespace.
+func ensureLedgerCredentials(ctx Context, stack *v1beta1.Stack) (keyID, secretName string, ready bool, err error) {
+	cred := &unstructured.Unstructured{}
+	cred.SetGroupVersionKind(ledgerCredentialsGVK)
+	cred.SetName("connectivity-" + stack.Name)
+	if _, err = controllerutil.CreateOrUpdate(ctx, ctx.GetClient(), cred, func() error {
+		if err := controllerutil.SetControllerReference(stack, cred, ctx.GetScheme()); err != nil {
```
NumaryBot, line 529: 🟠 **[major] Ledger credentials not cleaned up when Connectivity module is deleted**

When a Connectivity module is removed but the Stack remains, the cluster-scoped Credentials object (connectivity-<stack>) is kept alive because it is owned by the Stack rather than the Connectivity module. There is no Connectivity finalizer or cleanup path that deletes this object. That leaves the privileged ledger credential and the distributed secret registered even after the module is disabled or deleted, so module removal does not revoke the access it created.

**Suggestion:** Add a finalizer to the Connectivity resource that explicitly deletes the associated Credentials object (and any distributed secret) during deletion. Alternatively, set the Credentials object's owner reference to the Connectivity CR so garbage collection handles cleanup automatically.

NumaryBot (commented):
NumaryBot posted 3 new inline findings.

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

internal/resources/ledgers/exports.go
```diff
@@ -0,0 +1,38 @@
...
+	return isLedgerV3(version)
+}
+
+// V3GRPCBackendRef returns the connection details of the ledger v3 gRPC service
+// for the given stack: service name, port, and backend TLS material
+// (self-signed CA secret and SNI server name). It is the single source of
+// truth for how in-cluster clients reach the ledger v3 gRPC endpoint. The
+// default gRPC service port is assumed; stacks overriding the ledger Cluster
+// service port are not reachable through this helper.
+func V3GRPCBackendRef(stackName string) v1beta1.GatewayBackendRef {
```
NumaryBot: 🟠 **[major] Hardcoded default gRPC port ignores configured Ledger service gRPC port**

When a stack applies a LedgerConfiguration with spec.cluster.service.grpcPort set, the Ledger reconciler exposes the gateway backend using that configured port, but the port helper always passes 0 and therefore returns the default port for Connectivity. The delegated Connectivity then gets a spec.ledgerAddress pointing at the wrong service port and cannot reach Ledger on those stacks.

**Suggestion:** Pass the actual configured gRPC port (from the LedgerConfiguration) instead of 0 when constructing the ledger address used by the Connectivity module.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,509 @@
...
+	// Provision a god-mode ledger credential so connectivity-core can
+	// authenticate its gRPC calls. The ledger operator registers the public key
+	// on the ledger and distributes the private seed as a Secret in the stack
+	// namespace; connectivity-core is wired to it via spec.auth below.
+	authKeyID, authSecretName, credReady, err := ensureLedgerCredentials(ctx, stack)
+	if err != nil {
+		setCondition(connectivity, metav1.ConditionFalse, "LedgerCredentialsFailed", err.Error())
+		return err
+	}
+	if !credReady {
```
NumaryBot, line 195: 🟠 **[major] No watch on ledger Credentials — reconciler can stall indefinitely**

When the Connectivity reconciler creates the ledger Credentials object and returns a pending/application error waiting for status.phase=Ready, it will not be requeued automatically. The Credentials resource is not registered as a watched object; the reconciler only watches GatewayHTTPAPI, the delegated Connectivity, settings, and Ledger. On a normal install the controller can remain stuck in LedgerCredentialsPending until some unrelated watched object triggers a reconcile.

**Suggestion:** Add a Watch for the ledger Credentials resource in the controller setup so that a status update on that object (e.g. phase→Ready) triggers a reconcile of the owning Connectivity. Alternatively, use a periodic requeue (e.g. ctrl.Result{RequeueAfter: someInterval}) whenever the reconciler returns while waiting for Credentials readiness.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,509 @@
...
+// Stack rather than the namespaced Connectivity — a namespaced owner cannot own
+// a cluster-scoped resource. The ledger operator generates the Ed25519 keypair,
+// registers the public key on the Cluster matched by the stack selector, and
+// distributes the private seed as a Secret into the stack namespace.
+func ensureLedgerCredentials(ctx Context, stack *v1beta1.Stack) (keyID, secretName string, ready bool, err error) {
+	cred := &unstructured.Unstructured{}
+	cred.SetGroupVersionKind(ledgerCredentialsGVK)
+	cred.SetName("connectivity-" + stack.Name)
+	if _, err = controllerutil.CreateOrUpdate(ctx, ctx.GetClient(), cred, func() error {
+		if err := controllerutil.SetControllerReference(stack, cred, ctx.GetScheme()); err != nil {
```
NumaryBot, line 529: 🟠 **[major] Ledger credentials not cleaned up when Connectivity module is deleted**

When a Connectivity module is removed while the Stack remains, the cluster-scoped Credentials object (connectivity-<stack>) is kept alive because it is owned by the Stack rather than the Connectivity module. There is no Connectivity finalizer or cleanup path that deletes this object. That leaves the privileged ledger credential and the distributed secret registered even after the module is disabled or deleted, so module removal does not revoke the access it created.

**Suggestion:** Add a finalizer to the Connectivity resource that explicitly deletes the associated Credentials object (and any distributed secret) during deletion. Alternatively, set the Credentials object's owner reference to the Connectivity CR so garbage collection handles cleanup automatically.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,509 @@
...
+
+	// Resolve the ledger's effective version through the same path the module
+	// reconcilers use (module override, stack version, or the referenced
+	// Versions file), so the v3 gate also works for versionsFromFile stacks.
+	ledgerVersion, err := ResolveModuleVersion(ctx, stack, ledger)
+	if err != nil {
+		setCondition(connectivity, metav1.ConditionFalse, "LedgerVersionUnresolved", err.Error())
+		return NewPendingError().WithMessage("cannot resolve the ledger version: %s", err.Error())
+	}
+	if !ledgers.IsV3(ledgerVersion) {
```
NumaryBot, line 159: 🟠 **[major] Tear down delegated connectivity when ledger gate closes**

If a stack already provisioned Connectivity and the Ledger is later downgraded below v3 or becomes unready, the early returns only mark the module pending and leave the delegated connectivity.formance.com resource and GatewayHTTPAPI in place. The connectivity operator can therefore keep the workload running and exposed even though the prerequisite gate is no longer satisfied. The owned resources should be deleted or reconciled to a disabled state before returning from these gate failures.

**Suggestion:** Before returning from any Ledger gate failure (version check, readiness check), delete or disable the delegated Connectivity resource and GatewayHTTPAPI so the workload is not left running in an invalid state.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,509 @@
...
+		Group:   "ledger.formance.com",
+		Version: "v1alpha1",
+		Kind:    "Credentials",
+	}
+)
+
+var connectivityRequiredVerbs = []string{"get", "list", "watch", "create", "update", "patch", "delete"}
+
+//+kubebuilder:rbac:groups=formance.com,resources=connectivities,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=formance.com,resources=connectivities/status,verbs=get;update;patch
```
NumaryBot, line 100: 🟡 **[minor] Ledger version resolution fails for stacks using versionsFromFile**

For stacks that use .spec.versionsFromFile instead of .spec.version, the fallback never reads the Ledger entry from the referenced Versions object, so ledgerVersion stays empty and Connectivity remains stuck with LedgerNotV3 even when the Ledger module itself reconciled as v3 and is ready. Additionally, if the Versions object has no 'connectivity' key, reconciliation fails before reaching the capability/Ledger-v3 gate, even though the resolved version value is never actually used by the reconciler.

**Suggestion:** Use the same module-version resolution path as the Ledger reconciler to read the Ledger version before applying the v3 gate. Also make the connectivity version field optional or provide a sensible default (e.g. empty string) so the reconciler can proceed when no connectivity version is specified in versionsFromFile.

NumaryBot (commented):
NumaryBot posted 4 new inline findings.

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

flemzord (commented):
Reviewed as the root of the linked stack. The current head still contains the credential watch/cleanup, closed-gate teardown, configurable gRPC port, gateway TLS pending-state, and NetworkPolicy issues tracked in the existing threads; those fixes live in descendant PRs #501–#507 rather than in this head. I am not duplicating the inline findings or approving this root in isolation. It is safe to reconsider once the stack is merged/restacked so those fixes are part of the effective change.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,746 @@
...
+	// Provision a god-mode ledger credential so connectivity-core can
+	// authenticate its gRPC calls. The ledger operator registers the public key
+	// on the ledger and distributes the private seed as a Secret in the stack
+	// namespace; connectivity-core is wired to it via spec.auth below.
+	authKeyID, authSecretName, credReady, err := ensureLedgerCredentials(ctx, stack)
+	if err != nil {
+		setCondition(connectivity, metav1.ConditionFalse, "LedgerCredentialsFailed", err.Error())
+		return err
+	}
+	if !credReady {
```
NumaryBot, line 195: 🟠 **[major] No watch on ledger Credentials — reconciler can stall indefinitely**

When the Connectivity reconciler creates the ledger Credentials object and returns a pending/application error waiting for `status.phase=Ready`, it will not be requeued automatically because the Credentials resource is not registered as a watched object. The reconciler only watches GatewayHTTPAPI, the delegated Connectivity, settings, and Ledger. On a normal install the controller can remain stuck in `LedgerCredentialsPending` until some unrelated watched object triggers a reconcile.

**Suggestion:** Add a Watch for the ledger Credentials resource in the controller setup so that a status update on that object (e.g. phase→Ready) triggers a reconcile of the owning Connectivity. Alternatively, use a periodic requeue (e.g. `ctrl.Result{RequeueAfter: someInterval}`) whenever the reconciler returns while waiting for Credentials readiness.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,746 @@
...
+// Stack rather than the namespaced Connectivity — a namespaced owner cannot own
+// a cluster-scoped resource. The ledger operator generates the Ed25519 keypair,
+// registers the public key on the Cluster matched by the stack selector, and
+// distributes the private seed as a Secret into the stack namespace.
+func ensureLedgerCredentials(ctx Context, stack *v1beta1.Stack) (keyID, secretName string, ready bool, err error) {
+	cred := &unstructured.Unstructured{}
+	cred.SetGroupVersionKind(ledgerCredentialsGVK)
+	cred.SetName("connectivity-" + stack.Name)
+	if _, err = controllerutil.CreateOrUpdate(ctx, ctx.GetClient(), cred, func() error {
+		if err := controllerutil.SetControllerReference(stack, cred, ctx.GetScheme()); err != nil {
```
NumaryBot, line 529: 🟠 **[major] Ledger credentials not cleaned up when Connectivity module is deleted**

When a Connectivity module is removed but the Stack remains, the cluster-scoped Credentials object (`connectivity-<stack>`) is kept alive because it is owned by the Stack rather than the Connectivity module. There is no Connectivity finalizer or cleanup path that deletes this object, so the privileged ledger credential and distributed secret remain registered even after the module is disabled or deleted, leaving access that was provisioned but never revoked.

**Suggestion:** Add a finalizer to the Connectivity resource that explicitly deletes the associated Credentials object (and any distributed secret) during deletion. Alternatively, set the Credentials object's owner reference to the Connectivity CR so Kubernetes garbage collection handles cleanup automatically.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,746 @@
...
+		// delegated resources — doing so on every ledger readiness blip would flap
+		// the connectivity workload. Leave it in place and just report pending.
+		return NewPendingError().WithMessage("waiting for the ledger to be ready")
+	}
+
+	// Provision a god-mode ledger credential so connectivity-core can
+	// authenticate its gRPC calls. The ledger operator registers the public key
+	// on the ledger and distributes the private seed as a Secret in the stack
+	// namespace; connectivity-core is wired to it via spec.auth below.
+	authKeyID, authSecretName, credReady, err := ensureLedgerCredentials(ctx, stack)
```
NumaryBot, line 185: 🟠 **[major] Missing ledger Credentials CRD/RBAC causes hard failure instead of pending**

When the `ledger.formance.com` Credentials CRD or its RBAC is absent (e.g. an older ledger-operator install), `ensureLedgerCredentials` returns a hard NoMatch/Forbidden error and the module is marked failed. Since Credentials is an external capability needed only for Connectivity, this scenario should be treated as a pending/unavailable condition rather than a reconciliation failure.

**Suggestion:** Detect NoMatch and Forbidden errors from the Credentials API call and convert them into a `NewPendingError()` (or an explicit unavailable condition) so the module reports pending rather than failed when the CRD is not yet installed.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,746 @@
...
+	// service, port and backend TLS material (self-signed CA secret + SNI) that
+	// the gateway uses to reach the ledger. The port is resolved from the stack's
+	// LedgerConfiguration so a stack overriding spec.cluster.service.grpcPort
+	// stays reachable, consistently with the gateway backend.
+	backend, err := ledgers.V3GRPCBackendRef(ctx, stack.Name)
+	if err != nil {
+		setCondition(connectivity, metav1.ConditionFalse, "LedgerBackendResolveFailed", err.Error())
+		return err
+	}
+	ledgerAddress := fmt.Sprintf("%s:%d", backend.TLS.ServerName, backend.Port)
```
NumaryBot: 🟠 **[major] ledgerAddress uses SNI server name instead of the backend Service endpoint**

`ledgerAddress` is built from `backend.TLS.ServerName`, which is the SNI/certificate hostname. In clusters whose DNS domain is not `cluster.local`, `ledger-<stack>.<stack>.svc.cluster.local` may not resolve. The gateway upstream path uses `backend.Name` as the Service address and `TLS.ServerName` only for TLS verification; `ledgerAddress` should follow the same pattern.

**Suggestion:** Build `ledgerAddress` from the backend Service name/endpoint rather than `backend.TLS.ServerName`, and use `TLS.ServerName` only for the TLS SNI/verification field.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,746 @@
...
+		Kind:    "Connectivity",
+	}
+	connectivityAvailable bool
+
+	// ledgerCredentialsGVK is the cluster-scoped ledger.formance.com/Credentials
+	// resource. The connectivity module provisions a god-mode credential so
+	// connectivity-core can authenticate its gRPC calls to the stack's Ledger v3:
+	// the ledger operator generates the Ed25519 keypair, registers the public key
+	// on the matched ledger Cluster, and distributes the private seed as a Secret.
+	ledgerCredentialsGVK = schema.GroupVersionKind{
```
NumaryBot, line 83: 🟠 **[major] Ledger version resolved only from spec.version, not from versionsFromFile**

For stacks that use `.spec.versionsFromFile` instead of `.spec.version`, the fallback never reads the Ledger entry from the referenced `Versions` object, so `ledgerVersion` stays empty and Connectivity remains stuck with `LedgerNotV3` even when the Ledger module itself reconciled as v3 and is ready.

**Suggestion:** Use the same module-version resolution path as the Ledger reconciler (reading from the referenced Versions object when `versionsFromFile` is set) before applying the v3 gate.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,746 @@
...
+	stack, ok := strings.CutPrefix(object.GetName(), "connectivity-")
+	if !ok || stack == "" {
+		return "", false
+	}
+	return stack, true
+}
+
+func init() {
+	Init(
+		WithModuleReconciler(Reconcile,
```
NumaryBot: 🟡 **[minor] Version resolution fails on versionsFromFile stacks without a 'connectivity' entry**

For stacks that use `spec.versionsFromFile`, the module framework resolves the Connectivity version before entering Reconcile. If the referenced Versions object has no 'connectivity' key, reconciliation fails before reaching the capability/Ledger-v3 gate, even though the resolved version value is never actually used by the reconciler. This forces operators to add a meaningless version entry to their Versions file just to unblock reconciliation.

**Suggestion:** Make the version field optional or provide a sensible default (e.g. empty string) so the reconciler can proceed when no connectivity version is specified. If the version is genuinely unused, skip the version-resolution gate entirely for this module.

internal/resources/gateways/deployment.go
```diff
@@ -124,6 +125,12 @@ func configureBackendTLSVolumes(
...
 		secret := &v1.Secret{}
 		if err := ctx.GetClient().Get(ctx, core.GetNamespacedResourceName(namespace, secretName), secret); err != nil {
+			if apierrors.IsNotFound(err) {
+				// The backend module may not have provisioned its TLS Secret
+				// yet. Surface this as a pending (application) error so the
+				// framework retries instead of marking the Gateway as failed.
+				return core.NewPendingError().WithMessage("waiting for backend TLS Secret %s/%s", namespace, secretName)
+			}
 			return fmt.Errorf("getting backend TLS Secret %s/%s: %w", namespace, secretName, err)
 		}
```
NumaryBot, line 135: 🟡 **[minor] Missing backend TLS Secret treated as hard error instead of pending**

If the referenced backend TLS Secret doesn't exist yet (e.g., race with the owning module provisioning it), the code returns a hard error instead of a pending state — inconsistent with the 'broker not ready' handling in `internal/resources/gateways/init.go` which uses `NewPendingError()`. The labeled-secret watch will eventually retrigger reconciliation, but until then the Gateway will surface as errored rather than pending.

**Suggestion:** Check for `apierrors.IsNotFound(err)` and return `core.NewPendingError().WithMessage(...)` instead of a hard error when the Secret is not yet available.

NumaryBot (commented):
NumaryBot posted 7 new inline findings.

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

internal/resources/connectivities/init.go
Dav-14: C'est pas mon probleme

NumaryBot (commented):

Resolved 7 stale NumaryBot review threads (6 fixed, 1 outdated).

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

flemzord (commented):
I found two blocking lifecycle/capability issues and three controller-runtime correctness or policy-scope issues on the current head. CI is green, but the existing E2E suite does not exercise the Connectivity module.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,767 @@
...
+// Stack rather than the namespaced Connectivity — a namespaced owner cannot own
+// a cluster-scoped resource. The ledger operator generates the Ed25519 keypair,
+// registers the public key on the Cluster matched by the stack selector, and
+// distributes the private seed as a Secret into the stack namespace.
+func ensureLedgerCredentials(ctx Context, stack *v1beta1.Stack) (keyID, secretName string, ready bool, err error) {
+	cred := &unstructured.Unstructured{}
+	cred.SetGroupVersionKind(ledgerCredentialsGVK)
+	cred.SetName("connectivity-" + stack.Name)
+	if _, err = controllerutil.CreateOrUpdate(ctx, ctx.GetClient(), cred, func() error {
+		if err := controllerutil.SetControllerReference(stack, cred, ctx.GetScheme()); err != nil {
```
flemzord, line 529: This can leave a usable god-mode credential behind when a Stack is disabled. The shared disabled path only deletes types registered in `options.Owns` and does not run this reconciler or the Connectivity finalizer; this cluster-scoped Credentials object is owned by the Stack and is not in `Owns`. Please explicitly revoke/delete it on Stack disable and add a regression test for the disabled lifecycle.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,767 @@
...
+			"connectivity operator unavailable: connectivity.formance.com Connectivity CRD is not installed")
+		// The connectivity operator is gone, so we can neither provision nor
+		// manage the delegated workload — but if the ledger hard gate is also
+		// closed we must still tear down the god-mode Credentials and the gateway
+		// route we own, otherwise a closed gate leaves them behind just because
+		// this capability check short-circuits before the ledger gates below.
+		// Gated on the gate being closed so a transient operator outage (with a
+		// healthy ledger) does not flap those resources.
+		if ledgerGateClosed(ctx, stack) {
+			if err := teardownDelegated(ctx, stack, connectivity); err != nil {
```
flemzord: This can turn the advertised RBAC-unavailable pending state into a hard reconcile error. When `connectivityAvailable` is false because the SSAR denied access and the Ledger gate is closed, `teardownDelegated` still deletes the delegated CR; that delete returns `Forbidden`, while `ignoreAbsent` only ignores NoMatch/NotFound. Please keep this path pending while independently cleaning the route and credential, and cover the Forbidden case.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,767 @@
...
+		}
+		for _, version := range crd.Spec.Versions {
+			if version.Name != ledgerCredentialsGVK.Version || !version.Served {
+				continue
+			}
+			cred := &unstructured.Unstructured{}
+			cred.SetGroupVersionKind(ledgerCredentialsGVK)
+			// The raw builder callback receives a Context wrapping the manager;
+			// its client is long-lived, so it is safe to reuse for the mapping.
+			b.Watches(cred, handler.EnqueueRequestsFromMapFunc(func(_ context.Context, object client.Object) []reconcile.Request {
```
flemzord, line 738: This watch can prevent the controller from starting when the Credentials CRD exists but the ServiceAccount lacks `list` or `watch`. Presence alone registers an informer whose cache cannot sync, so controller-runtime eventually fails after `CacheSyncTimeout` instead of using the intended polling fallback. Please preflight accessibility/RBAC before registering the watch and fall back to the delayed requeue when unavailable.

internal/resources/connectivities/init.go
```diff
@@ -0,0 +1,767 @@
...
+		setCondition(connectivity, metav1.ConditionFalse, "LedgerBackendResolveFailed", err.Error())
+		return err
+	}
+	ledgerAddress := fmt.Sprintf("%s:%d", backend.Name, backend.Port)
+
+	object := &unstructured.Unstructured{}
+	object.SetGroupVersionKind(connectivityGVK)
+	object.SetNamespace(stack.Name)
+	object.SetName(connectivityDelegatedName)
+	if _, err := controllerutil.CreateOrUpdate(ctx, ctx.GetClient(), object, func() error {
```
flemzord: This can report Ready for a stale delegated spec after an update. `CreateOrUpdate` preserves the previous status and its `OperationResult` is ignored, so the later `status.phase == Ready` check may still describe the old image/TLS/auth/backend generation. Please return pending when this operation creates or updates the object (or verify an observed generation) and add an update-path test.

internal/resources/stacks/networkpolicies.go
```diff
@@ -182,6 +182,37 @@ func createNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
...
+		},
+		func(np *networkingv1.NetworkPolicy) error {
+			ledgerSelector := directLedgerV3Selector(stack.Name)
+			connectivity := connectivitySelector()
+			np.Spec = networkingv1.NetworkPolicySpec{
+				PodSelector: ledgerSelector,
+				PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
+				Ingress: []networkingv1.NetworkPolicyIngressRule{
+					{
+						From: []networkingv1.NetworkPolicyPeer{{PodSelector: &connectivity}},
```
flemzord: This creates an ingress allowance on every network-policy-enabled Stack, even without a Connectivity module, and permits every Ledger pod port from any same-namespace pod carrying only `app.kubernetes.io/name=connectivity`. Please scope policy creation to Stacks that actually have the module and narrow the source instance and destination port, with the necessary LedgerConfiguration watch to keep port overrides convergent.

internal/resources/connectivities/init.go
Dav-14: HUUUUm the thing is it will force the ledger to reconcile TWICE when reannabling a ledger

Disabling a stack should only disable the resource and not finalize ANYTHING

internal/resources/connectivities/init.go
Dav-14: Fixed in adecdab2. The CreateOrUpdate result is now checked: any create/update returns Pending before reading the preserved delegated status, so a stale Ready condition cannot leak through. Regression added for updating an already-Ready delegated resource.

internal/resources/connectivities/init.go
Dav-14: Fixed in adecdab2. Credentials watch registration now preflights a real List plus SSAR checks for list/watch. On denial the informer is skipped and delayed polling remains active. Tests cover both list and watch permission denial.

internal/resources/connectivities/init.go
Dav-14: Fixed in adecdab2. Stack-disabled reconciliation now runs explicit Connectivity credential cleanup before version resolution, and still attempts owned-resource teardown if cleanup fails. Regressions cover both an unresolvable version and partial cleanup failure.

internal/resources/connectivities/init.go
Dav-14: Fixed in adecdab2. When Connectivity RBAC is unavailable and the ledger gate is closed, reconciliation no longer deletes the inaccessible delegated CR. It independently removes GatewayHTTPAPI and Credentials, then remains pending. A Forbidden regression test covers this path.

internal/resources/stacks/networkpolicies.go
Dav-14: Fixed in adecdab2. The Connectivity NetworkPolicy is now gated on an active module, selects both name=connectivity and instance=connectivity, and allows only the configured Ledger v3 gRPC TCP port. Stack reconciliation also watches LedgerConfiguration changes. Unit, envtest, and Chainsaw coverage were added.

flemzord (dismissed):
All five review findings are addressed in adecdab2. The rerun is green across Kubernetes 1.31–1.35, including the previously failing 1.32 job. Approved.

internal/resources/connectivities/init.go
Dav-14: Fixed in `d9cc1cb7`. Disabling a Stack now only removes module-owned runtime objects and never invokes finalizers. The Connectivity Ledger Credentials are preserved across disable/re-enable, avoiding credential/key churn and the extra Ledger reconciliation. Actual Connectivity deletion still runs the `delete-ledger-credentials` finalizer. Regression tests cover both contracts.

Dav-14: Follow-up in `d9cc1cb7`: per Stack lifecycle semantics, `spec.disabled` no longer runs deletion finalizers. It removes module-owned runtime objects but retains the Ledger Credentials to avoid key churn and double Ledger reconciliation on re-enable. Actual Connectivity deletion still revokes the credential. Regression tests cover the disabled and deleted paths.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

internal/core/controllers.go
```diff
@@ -126,6 +126,13 @@ type ModuleController[T v1beta1.Module] func(ctx Context, stack *v1beta1.Stack,
 
 func ForModule[T v1beta1.Module](underlyingController ModuleController[T]) StackDependentObjectController[T] {
 	return func(ctx Context, stack *v1beta1.Stack, reconcilerOptions *ReconcilerOptions[T], t T) error {
+		if stack.Spec.Disabled {
```
NumaryBot: 🔴 **[blocker] Revoke module-managed credentials when disabling a stack**

When a provisioned Connectivity stack is disabled, this early return only removes objects listed in `Owns`; its cluster-scoped ledger `Credentials` is owned by the Stack and cleaned through the module finalizer, so it remains active with god-mode access after the workload is disabled. The disabled path needs to invoke equivalent module-specific cleanup for resources that cannot be module-owned.

NumaryBot: 🟠 **[major] Preserve the Stack owner reference before returning**

If a module is created while its Stack is already disabled, this return occurs before `SetOwnerReference` runs. The module then remains unowned, so deleting the Stack does not garbage-collect it; establish the owner reference before the disabled cleanup path.

Dav-14: This is intentional per the Stack lifecycle contract confirmed in this review: disabling a Stack pauses module runtime resources and must not execute deletion finalizers. Revoking this credential on disable causes key churn and a second Ledger reconciliation when the Stack is re-enabled. The credential remains owned by the Stack and is revoked by the Connectivity deletion finalizer on actual deletion. `TestDisabledStackKeepsLedgerCredentialsWithoutReconcilingModule` and `TestDeleteLedgerCredentialsFinalizerDeletesCredential` cover the two distinct paths.

Dav-14: Fixed in `4ac690ef`. The Stack owner reference is now established before the disabled early return, while version resolution remains deferred until after the disabled path. The regression test verifies that an initially disabled module keeps its Stack owner reference, skips finalizers, and removes its owned runtime resources.

NumaryBot (commented):

Resolved 2 stale NumaryBot review threads (0 fixed, 2 outdated).

Summary: https://github.com/formancehq/operator/pull/492#issuecomment-4992922939

flemzord (approved):
All review threads are addressed and resolved on 4ac690ef. The ownership fix is correct, the Stack disable lifecycle is intentional and covered by tests, and the full CI matrix is green across Kubernetes 1.31–1.35. Approved.

PR #493, opened 2026-07-20
feat(reconciliation): deploy dedicated v3 worker

flemzord:
## Summary

- Gate the split reconciliation topology on v3.0.0 and newer, including non-semver development images.
- Keep the API deployment on serve and add a dedicated reconciliation-worker deployment using worker, with a default replica count of one.
- Preserve the legacy single-deployment topology for older versions and wait for the API rollout before starting the worker during upgrades.
- Propagate Broker, topic mapping, circuit-breaker configuration, worker settings, and distinct OpenTelemetry service names to both workloads.
- Extend controller and Chainsaw coverage for the topology gate, worker arguments, settings, Broker environment, and Gateway isolation.

## Why

Reconciliation scheduling is now PostgreSQL-backed and must run outside the stateless API. The worker deployment can safely scale beyond one replica because durable claims, leases, fencing tokens, and advisory locks coordinate concurrent pods.

Depends on https://github.com/formancehq/reconciliation/pull/83

## Validation

- CGO_ENABLED=0 go test ./internal/resources/...
- Parsed all changed Chainsaw YAML resources successfully
- git diff --check

The Chainsaw scenario was updated but not executed locally because it requires a Kubernetes test cluster.


internal/resources/reconciliations/init.go
```diff
@@ -75,6 +110,10 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, reconciliation *v1beta1.Reconc
 	return nil
 }
 
+func usesV3Topology(version string) bool {
+	return !semver.IsValid(version) || semver.Compare(version, "v3.0.0") >= 0
```
NumaryBot, line 114: 🟠 **[major] Flavor suffixes cause valid v3.0.0 image tags to fail the semver gate**

For release tags such as `v3.0.0-scratch`, `semver.Compare(version, "v3.0.0")` treats `-scratch` as a pre-release identifier (per semver spec), making the comparison return -1. As a result, a `v3.0.0-scratch` image is incorrectly classified as pre-v3 and never receives the dedicated `reconciliation-worker` deployment, even though it is a fully supported v3 image flavor.

**Suggestion:** Strip known image-flavor suffixes (e.g., `-scratch`, `-alpine`, etc.) from the version string before running the semver gate, or use a separate mechanism to identify flavors so they are not confused with semver pre-release labels. For example: `cleanVersion := strings.SplitN(version, "-", 2)[0]` followed by appending the necessary semver build/pre-release only if the suffix is not a known flavor string.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/493#issuecomment-5045130319

NumaryBot (approved):

PR #497, opened 2026-07-24
fix(webhooks): subscribe to reconciliation events

flemzord:
## What changed

- add optional version-aware event publisher discovery
- register Reconciliation as an event publisher from `v3.0.0-0`
- include alpha, RC, stable, and non-SemVer development versions such as `main`
- keep Reconciliation v2 excluded from Webhooks subscriptions
- reconcile Webhooks when the Reconciliation module changes
- share the Reconciliation v3 threshold between topology and event publication

## Why

Reconciliation v3 publishes `reconciliation.alert.*` events to its standard message-bus topic. The Operator configured the Reconciliation publisher, but did not expose Reconciliation through the generic `EventPublisher` discovery used to build Webhooks consumers. As a result, Webhooks subscribed only to Ledger and Payments and could never receive Reconciliation events.

The version gate is optional so existing publishers such as Ledger, Payments, Orchestration, and TransactionPlane keep their current behavior.

## Validation

- generated manifests and objects successfully
- `go fmt ./...`
- `go vet ./...`
- `CGO_ENABLED=0 go test ./api/formance.com/v1beta1 ./internal/core ./internal/resources/reconciliations ./internal/resources/webhooks ./internal/resources/brokerconsumers`
- controller suite: 134/134 specs passed in serial envtest mode

The parallel all-suite run exposed an envtest concurrency flake in `internal/tests`; the isolated serial rerun passed all specs.

### DIFF
--- a/api/formance.com/v1beta1/reconciliation_types.go
+++ b/api/formance.com/v1beta1/reconciliation_types.go
@@ -17,8 +17,11 @@ limitations under the License.
 package v1beta1
 
 import (
+	"golang.org/x/mod/semver"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+const ReconciliationV3Version = "v3.0.0-0"
 
 type ReconciliationSpec struct {
 	StackDependency  `json:",inline"`
@@ -74,6 +77,16 @@ func (in *Reconciliation) GetVersion() string {
 	return in.Spec.Version
 }
 
+func (in Reconciliation) isEventPublisher() {}
+
+func (in Reconciliation) PublishesEvents(version string) bool {
+	return IsReconciliationV3(version)
+}
+
+func IsReconciliationV3(version string) bool {
+	return !semver.IsValid(version) || semver.Compare(version, ReconciliationV3Version) >= 0
+}
+
 func (a Reconciliation) IsDebug() bool {
 	return a.Spec.Debug
 }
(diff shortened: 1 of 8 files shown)


api/formance.com/v1beta1/reconciliation_types.go
```diff
@@ -74,6 +77,16 @@ func (in *Reconciliation) GetVersion() string {
...
 }
 
+func (in Reconciliation) isEventPublisher() {}
+
+func (in Reconciliation) PublishesEvents(version string) bool {
+	return IsReconciliationV3(version)
+}
+
+func IsReconciliationV3(version string) bool {
+	return !semver.IsValid(version) || semver.Compare(version, ReconciliationV3Version) >= 0
```
NumaryBot, line 87: 🟠 **[major] Unprefixed semver versions misclassified as non-SemVer/dev**

If a stack or Versions entry uses a semver tag without the leading `v` (e.g. `2.3.1`), `semver.IsValid` returns false and the code falls through to the non-SemVer/dev path, treating it as a development version. This causes Webhooks to subscribe to Reconciliation v2 even though the intent of the gate is to exclude v2. Other version-handling code in the codebase accepts unprefixed semver tags, so this gate is inconsistent and can produce incorrect behavior silently.

**Suggestion:** Normalize the version string before calling `semver.IsValid` — e.g. prepend `v` if it is absent (`if !strings.HasPrefix(ver, "v") { ver = "v" + ver }`). This ensures unprefixed semver tags are correctly parsed and compared rather than being classified as development/non-SemVer versions.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/497#issuecomment-5067393664

NumaryBot (approved):

PR #500, opened 2026-07-30
feat(connectivity): provision OpenTelemetry monitoring on the delegated resource

Dav-14:
## Summary

Wire OpenTelemetry observability into the Connectivity module, mirroring the Ledger v3 pattern.

- Resolve the stack's OTEL configuration via `settings.GetOpenTelemetryConfiguration(ctx, stack.Name, "connectivity")`. This is collector-aware: it points the signals at `otel-collector.<stack>:4318` when the per-stack collector Service exists, otherwise it honours the `opentelemetry.*` Settings.
- Embed the resolved config **inline** in the delegated resource's `spec.monitoring` (same shape as Ledger v3's `Cluster.spec.monitoring`). The connectivity operator's `MonitoringConfig` is a struct embedded in the CRD spec — there is **no separate `Monitoring` object** to create or reference. The connectivity operator turns `spec.monitoring` into `OTEL_*` env vars on the workload.
- Each present signal is enabled with the `otlp` exporter; traces are batched, metrics carry runtime instrumentation — matching the Ledger v3 mapping.

## Idempotency

The whole `spec.monitoring` block is rebuilt inside the existing `CreateOrUpdate` mutate closure on every reconcile, and pruned (`RemoveNestedField`) when telemetry is disabled — so repeated reconciles converge to the same object.

## Test plan

- [x] `go test ./internal/resources/connectivities/...` green
- [x] New unit tests: inline signal mapping, sorted resource attributes, idempotency (byte-for-byte identical spec on re-apply), omission of unconfigured signals, and pruning when disabled
- [x] New helpers at 100% coverage (`applyConnectivityMonitoring`, `connectivityMonitoringSpec`)
- [x] `just pre-commit` (tidy, lint, generate, manifests, helm, docs, settings-catalog)

## Notes

- Base branch is `feat/connectivity-module` — this stacks the observability wiring on top of the connectivity module work.
- End-to-end `OTEL_*` env-var emission is owned by the connectivity operator repo and covered there.

### DIFF
--- a/internal/resources/connectivities/init.go
+++ b/internal/resources/connectivities/init.go
@@ -287,10 +287,23 @@ func connectivityMonitoringSpec(configuration *settings.OpenTelemetryConfigurati
 	if len(configuration.Attributes) > 0 {
 		attributes := make([]string, 0, len(configuration.Attributes))
 		for key, value := range configuration.Attributes {
+			// GetOpenTelemetryConfiguration injects pod-name=$(POD_NAME), which
+			// only resolves when a downward-API POD_NAME env var is defined ahead
+			// of OTEL_RESOURCE_ATTRIBUTES on the workload. Unlike this operator's
+			// own env-var path (settings.otelEnvVars/collectorEnvVars), the
+			// connectivity operator emits OTEL_RESOURCE_ATTRIBUTES verbatim from
+			// spec.monitoring.attributes and defines no such env var, so any
+			// $(...) placeholder would surface literally in the telemetry. Forward
+			// only attributes with resolvable (literal) values.
+			if strings.Contains(value, "$(") {
+				continue
+			}
 			attributes = append(attributes, key+"="+value)
 		}
-		slices.Sort(attributes)
-		monitoring["attributes"] = strings.Join(attributes, ",")
+		if len(attributes) > 0 {
+			slices.Sort(attributes)
+			monitoring["attributes"] = strings.Join(attributes, ",")
+		}
 	}
 
 	signal := func(cfg *settings.OpenTelemetrySignalConfiguration, extra map[string]any) map[string]any {
--- a/internal/resources/connectivities/init.go
+++ b/internal/resources/connectivities/init.go
@@ -18,6 +18,10 @@ package connectivities
 
 import (
 	"fmt"
+	"maps"
+	"slices"
+	"strconv"
+	"strings"
 
 	authorizationv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -35,6 +39,7 @@ import (
 	"github.com/formancehq/operator/v3/internal/resources/gatewayhttpapis"
 	"github.com/formancehq/operator/v3/internal/resources/ledgers"
 	"github.com/formancehq/operator/v3/internal/resources/registries"
+	"github.com/formancehq/operator/v3/internal/resources/settings"
 )
 
 const (
@@ -148,6 +153,13 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
 		return err
 	}
 
+	// Collector-aware OTEL configuration, embedded inline in spec.monitoring.
+	monitoringConfiguration, err := settings.GetOpenTelemetryConfiguration(ctx, stack.Name, "connectivity")
+	if err != nil {
+		setCondition(connectivity, metav1.ConditionFalse, "MonitoringResolveFailed", err.Error())
+		return err
+	}
+
 	// Reuse the single source of truth for the ledger v3 gRPC connection: same
 	// service, port and backend TLS material (self-signed CA secret + SNI) that
 	// the gateway uses to reach the ledger.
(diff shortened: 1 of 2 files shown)


internal/resources/connectivities/init.go
```diff
@@ -248,6 +272,67 @@ func pullSecretsToUnstructured(secrets []corev1.LocalObjectReference) []any {
...
+	monitoring := map[string]any{
+		"serviceName": "connectivity",
+	}
+	if len(configuration.Attributes) > 0 {
+		attributes := make([]string, 0, len(configuration.Attributes))
+		for key, value := range configuration.Attributes {
+			attributes = append(attributes, key+"="+value)
+		}
+		slices.Sort(attributes)
+		monitoring["attributes"] = strings.Join(attributes, ",")
```
NumaryBot: 🟠 **[major] Define POD_NAME before using it in attributes**

When this config comes from `GetOpenTelemetryConfiguration`, `configuration.Attributes` always includes `pod-name=$(POD_NAME)`, but this delegated Connectivity spec only passes that string through `spec.monitoring.attributes`. The connectivity operator's monitoring env conversion emits `OTEL_RESOURCE_ATTRIBUTES` from this field and does not define a downward-API `POD_NAME` env var before it, so telemetry from Connectivity will report the literal `$(POD_NAME)` instead of the pod name. Either avoid forwarding this attribute for Connectivity or ensure the delegated workload defines `POD_NAME` before `OTEL_RESOURCE_ATTRIBUTES` is emitted.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/500#issuecomment-5131288516

internal/resources/connectivities/init.go
```diff
@@ -248,6 +263,61 @@ func pullSecretsToUnstructured(secrets []corev1.LocalObjectReference) []any {
...
```
NumaryBot: 🟠 **[major] POD_NAME env var not defined before use in OTEL resource attributes**

The `GetOpenTelemetryConfiguration` call populates `configuration.Attributes` with `pod-name=$(POD_NAME)`, and this value is forwarded verbatim via `spec.monitoring.attributes`. The connectivity operator's monitoring env conversion emits `OTEL_RESOURCE_ATTRIBUTES` from this field but does not define a downward-API `POD_NAME` environment variable ahead of it. As a result, telemetry from the delegated Connectivity workload reports the literal string `$(POD_NAME)` instead of the actual pod name, producing inaccurate telemetry metadata.

**Suggestion:** Either (a) strip the `pod-name` attribute before forwarding it to the delegated Connectivity spec, or (b) inject a `POD_NAME` environment variable sourced from the downward API (`fieldRef: fieldPath: metadata.name`) into the delegated workload's env block so that the shell-style substitution resolves correctly before `OTEL_RESOURCE_ATTRIBUTES` is evaluated.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/500#issuecomment-5131288516

internal/resources/connectivities/init.go
Dav-14: Fixed in 473b3144.

Verified the finding: `settings.GetOpenTelemetryConfiguration` always injects `pod-name=$(POD_NAME)`, and the connectivity operator (`formancehq/connectivity`, `misc/operator/.../monitoring_types.go` → `MonitoringEnvVars`) sets `OTEL_RESOURCE_ATTRIBUTES` verbatim from `spec.monitoring.attributes` with plain-value env vars and no downward-API `POD_NAME` — so the placeholder would surface literally.

Fix: `connectivityMonitoringSpec` now drops any attribute whose value carries an unresolvable `$(...)` placeholder before forwarding, keeping literal attributes (`stack` + custom), and omits `attributes` entirely if nothing resolvable remains. Chose stripping over option (b) since injecting `POD_NAME` would require a change in the connectivity operator repo, out of scope here. Tests updated; helpers remain at 100% coverage.

internal/resources/connectivities/init.go
Dav-14: Fixed in 473b3144.

Verified the finding: `settings.GetOpenTelemetryConfiguration` always injects `pod-name=$(POD_NAME)`, and the connectivity operator (`formancehq/connectivity`, `misc/operator/.../monitoring_types.go` → `MonitoringEnvVars`) sets `OTEL_RESOURCE_ATTRIBUTES` verbatim from `spec.monitoring.attributes` with plain-value env vars and no downward-API `POD_NAME` — so the placeholder would surface literally.

Fix: `connectivityMonitoringSpec` now drops any attribute whose value carries an unresolvable `$(...)` placeholder before forwarding, keeping literal attributes (`stack` + custom), and omits `attributes` entirely if nothing resolvable remains. Chose stripping over option (b) since injecting `POD_NAME` would require a change in the connectivity operator repo, out of scope here. Tests updated; helpers remain at 100% coverage.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/500#issuecomment-5131288516

PR #501, opened 2026-08-03
fix(connectivity): finalizer cleans up ledger credentials on deletion

Dav-14:

### DIFF
--- a/internal/resources/connectivities/init.go
+++ b/internal/resources/connectivities/init.go
@@ -26,6 +26,7 @@ import (
 	authorizationv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


internal/resources/connectivities/init.go
```diff
@@ -394,6 +394,30 @@ func ensureLedgerCredentials(ctx Context, stack *v1beta1.Stack) (keyID, secretNa
...
+// idempotent when the Credentials is already gone.
+//
+// This runs only on Connectivity CR deletion; stack disable is intentionally
+// left untouched (finalizers do not run on disable), so the credential is
+// retained for re-enable, consistent with Databases/BrokerTopics.
+func deleteLedgerCredentials(ctx Context, connectivity *v1beta1.Connectivity) error {
+	cred := &unstructured.Unstructured{}
+	cred.SetGroupVersionKind(ledgerCredentialsGVK)
+	cred.SetName("connectivity-" + connectivity.GetStack())
+	return client.IgnoreNotFound(ctx.GetClient().Delete(ctx, cred))
```
NumaryBot: 🟠 **[major] Finalizer can permanently block Connectivity deletion when Credentials CRD is absent**

When the `ledger.formance.com/Credentials` CRD is not installed (e.g., a failed or partial deployment, or a cluster where the CRD was later removed), the `Delete` call in the finalizer returns a RESTMapper/no-match error rather than a Kubernetes `NotFound` error. `client.IgnoreNotFound` does not suppress this kind of error, so the finalizer never completes and the Connectivity resource gets permanently stuck in a terminating state — even though there is no credential resource to clean up.

**Suggestion:** In the finalizer cleanup path, additionally check for `meta.IsNoMatchError` (from `k8s.io/apimachinery/pkg/api/meta`) and `runtime.IsNotRegisteredError` to detect missing CRDs/API groups, and treat those the same as a `NotFound` — i.e., proceed with removing the finalizer. For example:
```go
if err := r.client.Delete(ctx, creds); err != nil {
    if !apierrors.IsNotFound(err) && !meta.IsNoMatchError(err) {
        return err
    }
}
```
This keeps the finalizer idempotent regardless of whether the dependent API is registered.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/501#issuecomment-5164016279

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/501#issuecomment-5164016279

flemzord (approved):
Looks good to me.

PR #502, opened 2026-08-03
fix(gateway): treat missing backend TLS secret as pending

Dav-14:
When a backend module has not yet provisioned its TLS Secret, the Gateway
deployment reconciler fetched the Secret and returned any error as a hard
failure. A NotFound during this race surfaced the Gateway as errored rather
than pending.

Return core.NewPendingError() on apierrors.IsNotFound so the framework treats
it as pending (matching how other 'not ready yet' conditions are handled) and
retries. All other Get errors remain hard errors.

---

### DIFF
--- a/internal/resources/gateways/deployment.go
+++ b/internal/resources/gateways/deployment.go
@@ -9,6 +9,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/formancehq/operator/v3/api/formance.com/v1beta1"
 	"github.com/formancehq/operator/v3/internal/core"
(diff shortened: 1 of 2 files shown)


internal/resources/gateways/deployment.go
```diff
@@ -124,6 +125,12 @@ func configureBackendTLSVolumes(
 	for _, secretName := range sortedSecretNames {
 		secret := &v1.Secret{}
 		if err := ctx.GetClient().Get(ctx, core.GetNamespacedResourceName(namespace, secretName), secret); err != nil {
+			if apierrors.IsNotFound(err) {
+				// The backend module may not have provisioned its TLS Secret
+				// yet. Surface this as a pending (application) error so the
+				// framework retries instead of marking the Gateway as failed.
+				return core.NewPendingError().WithMessage("waiting for backend TLS Secret %s/%s", namespace, secretName)
+			}
```
Dav-14, line 133: will it retry correctly ? does it watch correctly ?

Dav-14: Verified — both correct, via the Secret watch (not via requeue):

- **Watch:** the Gateway watches Secrets labeled `formance.com/gateway-backend-tls=true` and maps them to the Gateway in the same stack (`internal/resources/gateways/init.go:181`). The ledger v3 backend cert — which connectivity reuses — is created with exactly that label (`internal/resources/ledgers/v3_tls.go:140`), so a create/update of that Secret re-enqueues this Gateway.
- **Retry:** `NewPendingError()` returns `ctrl.Result{}, nil` (no `RequeueAfter`), so it does not self-requeue — it relies on that watch. When the backend module provisions the labeled Secret, the watch fires → reconcile re-runs → the Secret is found and the pending clears.

Caveat: retry is now purely watch-driven (a hard error would have retried via controller-runtime backoff). Correct as long as every backend TLS Secret carries `GatewayBackendTLSSecretLabel`, which the platform's backends do. Resolving.

flemzord (approved):
Looks good to me.

PR #504, opened 2026-08-03
fix(connectivity): watch ledger Credentials so pending reconcile is retriggered

Dav-14:
The Connectivity reconciler returns a PendingError while the cluster-scoped
ledger Credentials it provisions (connectivity-<stack>) is not yet Ready
(LedgerCredentialsPending). The reconcile loop treats a PendingError as a
terminal ctrl.Result{} with no RequeueAfter, and the Credentials is owned by
the Stack rather than the namespaced Connectivity, so nothing re-triggered the
module when the ledger operator flipped the Credentials status.phase to Ready:
the reconcile could stall indefinitely.

Register a raw builder watch on the ledger.formance.com/v1alpha1 Credentials
GVK (unstructured, mirroring withConnectivityClusterWatch) that maps a
Credentials event back to the Connectivity in the matching stack, derived from
the connectivity-<stack> name and listed via the stack field index. The watch
is gated on the Credentials CRD being installed so controller setup never fails
when the ledger operator is absent. RBAC already grants watch on credentials.

Add unit tests for the mapping (enqueues the matching Connectivity, returns
nothing for a stack without one, ignores foreign Credentials) and for the
capability gate (disabled when the CRD is absent or discovery fails).

---

### DIFF
--- a/internal/resources/connectivities/init.go
+++ b/internal/resources/connectivities/init.go
@@ -17,6 +17,7 @@ limitations under the License.
 package connectivities
 
 import (
+	"context"
 	"fmt"
 	"maps"
 	"slices"
@@ -33,7 +34,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/formancehq/operator/v3/api/formance.com/v1beta1"
 	. "github.com/formancehq/operator/v3/internal/core"
@@ -73,6 +76,13 @@ var (
 		Version: "v1alpha1",
 		Kind:    "Credentials",
 	}
+
+	// ledgerCredentialsWatchAvailable records whether, at controller start-up,
+	// the ledger Credentials CRD was present so the reconciler could register a
+	// watch on it. When true, changes to the connectivity-<stack> Credentials
+	// (notably its status.phase flipping to Ready) re-trigger the owning
+	// Connectivity's reconcile.
+	ledgerCredentialsWatchAvailable bool
 )
 
 var connectivityRequiredVerbs = []string{"get", "list", "watch", "create", "update", "patch", "delete"}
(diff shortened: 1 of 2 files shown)


internal/resources/connectivities/init.go
```diff
@@ -512,12 +522,106 @@ func canAccessConnectivityResource(ctx Context, gvk schema.GroupVersionKind, res
...
+// watch needs. The watch is gated on the Credentials CRD being installed so
+// controller setup never fails when the ledger operator is absent.
+func withLedgerCredentialsWatch() ReconcilerOption[*v1beta1.Connectivity] {
+	return func(options *ReconcilerOptions[*v1beta1.Connectivity]) {
+		options.Raws = append(options.Raws, func(ctx Context, b *builder.Builder) error {
+			crds := &apiextensionsv1.CustomResourceDefinitionList{}
+			if err := ctx.GetAPIReader().List(ctx, crds); err != nil {
+				ledgerCredentialsWatchAvailable = false
+				log.FromContext(ctx).Info("ledger Credentials watch unavailable; continuing without it", "error", err)
+				return nil
```
flemzord, line 550: This can leave Connectivity stuck indefinitely when the Credentials CRD is installed after this controller starts. Setup continues without registering the watch, and ledgerCredentialsWatchAvailable is never used; once credential creation starts succeeding, Reconcile returns a non-requeuing PendingError while no event can wake it. Consider failing setup until the watch can be registered, or adding a retry fallback whenever the watch is unavailable, with a test covering CRD installation after startup.

flemzord (changes requested):
One blocking reconciliation issue needs to be addressed.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/504#issuecomment-5167371762

internal/resources/connectivities/init.go
```diff
@@ -512,12 +522,106 @@ func canAccessConnectivityResource(ctx Context, gvk schema.GroupVersionKind, res
...
+			// The raw builder callback receives a Context wrapping the manager;
+			// its client is long-lived, so it is safe to reuse for the mapping.
+			b.Watches(cred, handler.EnqueueRequestsFromMapFunc(func(_ context.Context, object client.Object) []reconcile.Request {
+				return mapLedgerCredentialsToConnectivity(ctx, object)
+			}))
+			log.FromContext(ctx).Info("ledger Credentials watch registered", "gvk", ledgerCredentialsGVK)
+			return true
+		}
+	}
+	log.FromContext(ctx).Info("ledger Credentials CRD is not available; connectivity will not watch credentials", "gvk", ledgerCredentialsGVK)
```
NumaryBot, line 578: 🟠 **[major] Connectivity can stall indefinitely when Credentials CRD is installed after controller startup**
<sub>reported by codex, flemzord</sub>

When the ledger Credentials CRD is not present at controller startup, the watch is skipped and `ledgerCredentialsWatchAvailable` is never set to true. If the CRD is installed later (e.g., ledger operator deployed after this controller), the reconciler can create the Credentials object and return a non-requeuing PendingError while waiting for it to become Ready — but no informer exists to deliver the status-change event. The Connectivity object then remains stuck indefinitely with no trigger to resume reconciliation.

**Suggestion:** Add a polling/requeue fallback for when the watch is unavailable: either periodically retry registering the watch after CRD discovery fails, or ensure the reconciler requeues with a reasonable delay (instead of returning a non-requeuing PendingError) when `ledgerCredentialsWatchAvailable` is false so the missing watch cannot cause a permanent stall.

internal/resources/connectivities/init.go
Dav-14: Update: I reverted the requeue change (a121cf39) — it didn't belong in `core`.

On the "CRD installed after this controller starts" case: this is a controller-runtime constraint rather than something specific to this watch. `builder.Watches`/`Owns` informers must be registered before `mgr.Start()`, so a watch on `ledger.formance.com/Credentials` can only exist when that CRD is present at start-up — it can't be added afterwards. An owner-reference approach doesn't change that: it registers the same informer at setup, and a namespaced `Connectivity` can't own the cluster-scoped `Credentials` anyway (that's why it's Stack-owned), so `EnqueueRequestForOwner` isn't available here either.

This is consistent with every other capability gate in the operator — `withConnectivityClusterWatch`, `ledgerV3ClusterAvailable`, `ledgerV3CertManagerAvailable` all register at start-up, degrade when the dependency CRD is absent, and pick it up on the next controller restart. Installing a dependency operator's CRDs is a deploy-time/restart event in our model, so in practice the Credentials watch is registered in the nominal case.

If you'd rather close the after-startup gap explicitly, the only CRD-independent trigger is the distributed seed Secret (core/v1) — I can add a scoped watch on it as a follow-up. Happy to do that if you think it's worth diverging from the established gate pattern for this module.

internal/resources/connectivities/init.go
Dav-14: Intended behaviour — not changing this here.

Registering the watch requires the `ledger.formance.com/Credentials` informer at controller start-up: `builder.Watches` must run before `mgr.Start()`, and controller-runtime can't add an informer afterwards. When a dependency operator's CRDs are rolled out later, the operator is rebooted as part of that rollout — a planned step — which re-registers the watch. This is exactly how every other capability gate here behaves (`withConnectivityClusterWatch`, `ledgerV3ClusterAvailable`, `ledgerV3CertManagerAvailable`): registered at start-up, degraded when the dependency CRD is absent, picked up on the next restart.

The suggested requeue/poll fallback was implemented (a121cf39) and reverted: it meant plumbing a requeue delay through `core` for a case an operator reboot already covers by design. If we ever want to cover it without a reboot, the only CRD-independent trigger is the distributed seed Secret (core/v1) — a deliberate divergence from the gate pattern, not needed here.

internal/resources/connectivities/init.go
Dav-14: Decision: intended behaviour. The after-startup case is covered by an operator reboot on dependency-CRD rollout (a planned step), consistent with the other capability gates, so I'm leaving the CR watch as-is. Full rationale in the NumaryBot thread on this file.

flemzord (approved):

PR #505, opened 2026-08-03
fix(connectivity): tear down delegated resources when the ledger gate closes

Dav-14:
When a stack had already provisioned the delegated Connectivity and its
GatewayHTTPAPI, closing a hard Ledger gate on a later reconcile only set a
pending condition and returned; the delegated workload kept running and stayed
exposed through the gateway even though its prerequisite no longer held.

Introduce teardownDelegated(ctx, stack, connectivity), which idempotently
deletes both the delegated Connectivity and the GatewayHTTPAPI
(client.IgnoreNotFound). Call it from the hard/persistent gates only:

- LedgerNotFound  (module removed)      -> teardown
- LedgerNotV3     (real downgrade)      -> teardown

Leave the transient gates untouched so a momentary blip does not flap the
workload:

- LedgerVersionUnresolved (resolution error, not a downgrade) -> keep
- LedgerNotReady          (v3 but momentarily not ready)      -> keep

Add unit tests covering both the teardown-on-not-v3 and keep-on-v3-not-ready
paths.

---

### DIFF
--- a/internal/resources/connectivities/init.go
+++ b/internal/resources/connectivities/init.go
@@ -100,6 +100,18 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
 	if !connectivityAvailable {
 		setCondition(connectivity, metav1.ConditionFalse, "OperatorUnavailable",
 			"connectivity operator unavailable: connectivity.formance.com Connectivity CRD is not installed")
+		// The connectivity operator is gone, so we can neither provision nor
+		// manage the delegated workload — but if the ledger hard gate is also
+		// closed we must still tear down the god-mode Credentials and the gateway
+		// route we own, otherwise a closed gate leaves them behind just because
+		// this capability check short-circuits before the ledger gates below.
+		// Gated on the gate being closed so a transient operator outage (with a
+		// healthy ledger) does not flap those resources.
+		if ledgerGateClosed(ctx, stack) {
+			if err := teardownDelegated(ctx, stack, connectivity); err != nil {
+				return err
+			}
+		}
 		return NewPendingError().WithMessage("connectivity operator unavailable: connectivity.formance.com Connectivity CRD is not installed")
 	}
 
@@ -312,10 +324,42 @@ func teardownDelegated(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.
 	// Credentials also cascades the ledger operator's key deregistration and the
 	// distributed private-key Secret, which stack-namespace GC never reclaims.
 	return errors.Join(
-		client.IgnoreNotFound(ctx.GetClient().Delete(ctx, delegated)),
-		client.IgnoreNotFound(ctx.GetClient().Delete(ctx, httpAPI)),
+		ignoreAbsent(ctx.GetClient().Delete(ctx, delegated)),
+		ignoreAbsent(ctx.GetClient().Delete(ctx, httpAPI)),
 		deleteLedgerCredentials(ctx, connectivity),
 	)
+}
+
+// ignoreAbsent drops NotFound and NoMatch (the whole CRD is not installed)
+// errors, so a teardown delete is a no-op when the resource — or its API — is
+// already gone. This matters on the capability-unavailable path, where the
+// connectivity CRD may have been removed after startup.
+func ignoreAbsent(err error) error {
+	if apimeta.IsNoMatchError(err) {
+		return nil
+	}
+	return client.IgnoreNotFound(err)
+}
+
+// ledgerGateClosed reports whether the ledger prerequisite is definitively gone
+// — the module was removed, or resolves to a non-v3 version — which are the hard
+// gates that warrant tearing down the delegated resources. It mirrors the gate
+// decisions in Reconcile and deliberately returns false on transient states (an
+// unresolvable version, a not-yet-ready ledger) and on any lookup error, so the
+// workload is never flapped on a blip.
+func ledgerGateClosed(ctx Context, stack *v1beta1.Stack) bool {
+	ledger, err := getStackLedger(ctx, stack.Name)
+	if err != nil {
+		return false
+	}
+	if ledger == nil {
+		return true
+	}
+	ledgerVersion, err := ResolveModuleVersion(ctx, stack, ledger)
+	if err != nil {
+		return false
+	}
+	return !ledgers.IsV3(ledgerVersion)
 }
 
 // connectivityAPIBackendRef points the gateway at the connectivity-api Service
--- a/internal/resources/connectivities/init.go
+++ b/internal/resources/connectivities/init.go
@@ -111,6 +111,14 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
 	}
 	if ledger == nil {
 		setCondition(connectivity, metav1.ConditionFalse, "LedgerNotFound", "stack has no Ledger module")
+		// Hard gate: the Ledger module has been removed from the stack. This is a
+		// definitive loss of the prerequisite (not a momentary blip), so tear down
+		// any previously provisioned delegated Connectivity + GatewayHTTPAPI rather
+		// than leaving the workload running and exposed through the gateway with no
+		// ledger to ingest into.
+		if err := teardownDelegated(ctx, stack, connectivity); err != nil {
+			return err
+		}
 		return NewPendingError().WithMessage("connectivity requires a Ledger module on the stack")
 	}
 
@@ -120,15 +128,32 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
 	ledgerVersion, err := ResolveModuleVersion(ctx, stack, ledger)
 	if err != nil {
 		setCondition(connectivity, metav1.ConditionFalse, "LedgerVersionUnresolved", err.Error())
+		// Transient: we could not resolve the ledger version (e.g. a referenced
+		// Versions file not yet present). This is an error *resolving* the
+		// prerequisite, not a definitive downgrade below v3, so we deliberately do
+		// NOT tear down the delegated resources here — flapping the workload on a
+		// transient resolution hiccup would be worse than briefly keeping it up.
 		return NewPendingError().WithMessage("cannot resolve the ledger version: %s", err.Error())
 	}
 	if !ledgers.IsV3(ledgerVersion) {
 		setCondition(connectivity, metav1.ConditionFalse, "LedgerNotV3",
 			fmt.Sprintf("connectivity requires a Ledger v3 (found %q)", ledgerVersion))
+		// Hard gate: the ledger version resolved but is not v3 (i.e. a real
+		// downgrade). Connectivity binds to the ledger v3 gRPC surface, so the
+		// prerequisite no longer holds — tear down the delegated Connectivity +
+		// GatewayHTTPAPI before returning pending.
+		if err := teardownDelegated(ctx, stack, connectivity); err != nil {
+			return err
+		}
 		return NewPendingError().WithMessage("connectivity requires a Ledger v3")
 	}
 	if !ledger.IsReady() {
 		setCondition(connectivity, metav1.ConditionFalse, "LedgerNotReady", "waiting for the ledger to be ready")
+		// Transient: the ledger exists and is v3 but is momentarily not ready.
+		// Unlike the hard gates above (LedgerNotFound / LedgerNotV3), the
+		// prerequisite still holds, so we deliberately do NOT tear down the
+		// delegated resources — doing so on every ledger readiness blip would flap
+		// the connectivity workload. Leave it in place and just report pending.
 		return NewPendingError().WithMessage("waiting for the ledger to be ready")
 	}
 
@@ -253,6 +278,34 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
 
 	setCondition(connectivity, metav1.ConditionTrue, "Ready", "Connectivity is ready")
 	return nil
+}
+
+// teardownDelegated deletes the delegated Connectivity resource and the
+// GatewayHTTPAPI provisioned for the stack. It is invoked when a hard/persistent
+// ledger gate closes (the Ledger module was removed, or its version resolved to
+// something below v3) so the connectivity workload is not left running and
+// exposed through the gateway once its prerequisite no longer holds. There is no
+// automatic pruning of these module-owned objects on a normal reconcile
+// early-return — controller-runtime garbage collection only removes them when
+// the Connectivity CR itself is deleted — so the teardown must be explicit.
+//
+// Deletes use client.IgnoreNotFound so the helper is idempotent and safe to call
+// on every reconcile while the gate stays closed (including when the resources
+// were never created). The names/scopes mirror how they are provisioned in
+// Reconcile: the delegated Connectivity is namespaced (name == namespace ==
+// stack name) and the GatewayHTTPAPI is cluster-scoped ("<stack>-connectivity").
+func teardownDelegated(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connectivity) error {
+	delegated := &unstructured.Unstructured{}
+	delegated.SetGroupVersionKind(connectivityGVK)
+	delegated.SetNamespace(stack.Name)
+	delegated.SetName(stack.Name)
+	if err := client.IgnoreNotFound(ctx.GetClient().Delete(ctx, delegated)); err != nil {
+		return err
+	}
+
+	httpAPI := &v1beta1.GatewayHTTPAPI{}
+	httpAPI.SetName(GetObjectName(connectivity.GetStack(), LowerCaseKind(ctx, connectivity)))
+	return client.IgnoreNotFound(ctx.GetClient().Delete(ctx, httpAPI))
 }
 
 // connectivityAPIBackendRef points the gateway at the connectivity-api Service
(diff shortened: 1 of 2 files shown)


internal/resources/connectivities/init.go
```diff
@@ -255,6 +280,34 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
...
+	delegated.SetGroupVersionKind(connectivityGVK)
+	delegated.SetNamespace(stack.Name)
+	delegated.SetName(stack.Name)
+	if err := client.IgnoreNotFound(ctx.GetClient().Delete(ctx, delegated)); err != nil {
+		return err
+	}
+
+	httpAPI := &v1beta1.GatewayHTTPAPI{}
+	httpAPI.SetName(GetObjectName(connectivity.GetStack(), LowerCaseKind(ctx, connectivity)))
+	return client.IgnoreNotFound(ctx.GetClient().Delete(ctx, httpAPI))
```
flemzord: This leaves the cluster-scoped god-mode Credentials and its distributed private-key Secret behind when a hard Ledger gate closes. additionalNamespaces always includes the stack namespace, so removing the Ledger does not garbage-collect that Secret; during a blocked v3-to-v2 migration the public key may also remain registered on the existing v3 Cluster. Consider deleting the Connectivity-owned Credentials as part of this hard teardown and add assertions for both the Credentials and replica Secret.

flemzord (changes requested):
One blocking credential-lifecycle issue needs to be addressed.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/505#issuecomment-5168147906

internal/resources/connectivities/init.go
```diff
@@ -278,6 +303,34 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
...
+// on every reconcile while the gate stays closed (including when the resources
+// were never created). The names/scopes mirror how they are provisioned in
+// Reconcile: the delegated Connectivity is namespaced (name == namespace ==
+// stack name) and the GatewayHTTPAPI is cluster-scoped ("<stack>-connectivity").
+func teardownDelegated(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connectivity) error {
+	delegated := &unstructured.Unstructured{}
+	delegated.SetGroupVersionKind(connectivityGVK)
+	delegated.SetNamespace(stack.Name)
+	delegated.SetName(stack.Name)
+	if err := client.IgnoreNotFound(ctx.GetClient().Delete(ctx, delegated)); err != nil {
```
NumaryBot: 🟠 **[major] Gateway route left exposed when delegated Connectivity deletion fails**

When a hard ledger gate closes and deleting the delegated Connectivity returns a non-NotFound error (e.g. the connectivity.formance.com CRD/API was removed after startup, making connectivityAvailable stale), the early return exits before attempting to delete the cluster-scoped GatewayHTTPAPI. This leaves the public route in place even though it could be removed independently, undermining the exposure-reduction goal of the hard teardown.

**Suggestion:** Collect the error from the delegated Connectivity deletion rather than returning immediately, then proceed to delete the GatewayHTTPAPI unconditionally, and only return a combined/first error afterwards. This ensures both resources are torn down regardless of which deletion fails.

internal/resources/connectivities/init.go
```diff
@@ -278,6 +303,34 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
...
+		return err
+	}
+
+	httpAPI := &v1beta1.GatewayHTTPAPI{}
+	httpAPI.SetName(GetObjectName(connectivity.GetStack(), LowerCaseKind(ctx, connectivity)))
+	return client.IgnoreNotFound(ctx.GetClient().Delete(ctx, httpAPI))
```
NumaryBot: 🟠 **[major] Cluster-scoped Credentials and replica Secret not cleaned up on hard teardown**

When a hard Ledger gate closes, the cluster-scoped Credentials object and its associated distributed private-key Secret are not deleted. Because additionalNamespaces always includes the stack namespace, removing the Ledger does not trigger garbage collection of that Secret. During a blocked v3-to-v2 migration the public key may also remain registered on the existing v3 Cluster, leaving stale credentials behind.

**Suggestion:** Include deletion of the Connectivity-owned Credentials (and their replica Secret) as part of the hard teardown path, and add test assertions covering both the Credentials object and the replica Secret to prevent regressions.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/505#issuecomment-5168147906

internal/resources/connectivities/init.go
```diff
@@ -255,6 +280,34 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
...
+// on every reconcile while the gate stays closed (including when the resources
+// were never created). The names/scopes mirror how they are provisioned in
+// Reconcile: the delegated Connectivity is namespaced (name == namespace ==
+// stack name) and the GatewayHTTPAPI is cluster-scoped ("<stack>-connectivity").
+func teardownDelegated(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connectivity) error {
+	delegated := &unstructured.Unstructured{}
+	delegated.SetGroupVersionKind(connectivityGVK)
+	delegated.SetNamespace(stack.Name)
+	delegated.SetName(stack.Name)
+	if err := client.IgnoreNotFound(ctx.GetClient().Delete(ctx, delegated)); err != nil {
```
NumaryBot: 🟠 **[major] Gateway route left exposed when delegated Connectivity deletion fails**

When a hard ledger gate closes and deleting the delegated Connectivity returns a non-NotFound error, the early return exits before attempting to delete the cluster-scoped GatewayHTTPAPI. This leaves the public route in place even though it could be removed independently, undermining the exposure-reduction goal of the hard teardown.

**Suggestion:** Collect the error from the delegated Connectivity deletion rather than returning immediately, then proceed to delete the GatewayHTTPAPI unconditionally, and only return a combined/first error afterwards. This ensures both resources are torn down regardless of which deletion fails.

internal/resources/connectivities/init.go
NumaryBot: 🟠 **[major] Cluster-scoped Credentials and replica Secret not cleaned up on hard teardown**
<sub>reported by flemzord, NumaryBot</sub>

When a hard Ledger gate closes, the cluster-scoped Credentials object and its associated distributed private-key Secret are not deleted. Because additionalNamespaces always includes the stack namespace, removing the Ledger does not trigger garbage collection of that Secret. During a blocked v3-to-v2 migration the public key may also remain registered on the existing v3 Cluster, leaving stale credentials behind.

**Suggestion:** Include deletion of the Connectivity-owned Credentials (and their replica Secret) as part of the hard teardown path, and add test assertions covering both the Credentials object and the replica Secret to prevent regressions.

Dav-14: Fixed in fd721a5f. `teardownDelegated` now deletes the cluster-scoped Credentials via `deleteLedgerCredentials` (which cascades the ledger operator's key deregistration and the distributed private-key Secret that stack-namespace GC never reclaims), so the god-mode credential is cleaned up on a hard gate close. It also attempts every deletion independently (`errors.Join`) so a failed delegated-Connectivity delete no longer leaves the route or the Credentials behind. Tests assert Credentials cleanup on hard teardown, retention on a transient gate, and that every deletion is attempted when one fails.

internal/resources/connectivities/init.go
Dav-14: Fixed in fd721a5f: `teardownDelegated` now joins the errors from all deletions (`errors.Join`) instead of returning on the first, so the GatewayHTTPAPI route is torn down even when the delegated Connectivity delete fails. Covered by `TestTeardownDelegatedAttemptsEveryDeletion`.

internal/resources/connectivities/init.go
Dav-14: Fixed in fd721a5f: the hard teardown now deletes the cluster-scoped Credentials (cascading the ledger operator's key deregistration and the distributed Secret). Tests assert cleanup on teardown and retention on a transient gate.

internal/resources/connectivities/init.go
```diff
@@ -255,6 +281,43 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
...
+// Deletes use client.IgnoreNotFound so the helper is idempotent and safe to call
+// on every reconcile while the gate stays closed (including when the resources
+// were never created). The names/scopes mirror how they are provisioned in
+// Reconcile: the delegated Connectivity is namespaced (name == namespace ==
+// stack name) and the GatewayHTTPAPI is cluster-scoped ("<stack>-connectivity").
+func teardownDelegated(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connectivity) error {
+	delegated := &unstructured.Unstructured{}
+	delegated.SetGroupVersionKind(connectivityGVK)
+	delegated.SetNamespace(stack.Name)
+	delegated.SetName(stack.Name)
```
NumaryBot, line 314: 🟠 **[major] Gateway route left exposed when delegated Connectivity deletion fails**

When a hard ledger gate closes and deleting the delegated Connectivity returns a non-NotFound error, the early return exits before attempting to delete the cluster-scoped GatewayHTTPAPI. This leaves the public route active even though it could be removed independently, undermining the exposure-reduction goal of the hard teardown.

**Suggestion:** Collect the error from the delegated Connectivity deletion rather than returning immediately, then proceed to delete the GatewayHTTPAPI unconditionally, and only return a combined/first error afterwards. This ensures both resources are torn down regardless of which individual deletion fails.

internal/resources/connectivities/init.go
```diff
@@ -255,6 +281,43 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
...
+
+	httpAPI := &v1beta1.GatewayHTTPAPI{}
+	httpAPI.SetName(GetObjectName(connectivity.GetStack(), LowerCaseKind(ctx, connectivity)))
+
+	// Attempt every deletion independently rather than bailing on the first
+	// error: a failure to delete one resource must not leave the public gateway
```
NumaryBot, line 320: 🟠 **[major] Cluster-scoped Credentials and replica Secret not cleaned up on hard teardown**
<sub>reported by flemzord, NumaryBot</sub>

When a hard ledger gate closes, the cluster-scoped Credentials object and its associated distributed private-key Secret are not deleted as part of the teardown path. Because additionalNamespaces always includes the stack namespace, removing the Ledger does not trigger garbage collection of that Secret. During a blocked v3-to-v2 migration the public key may also remain registered on the existing v3 Cluster, leaving stale credentials behind after the gate closes.

**Suggestion:** Include deletion of the Connectivity-owned Credentials (and their replica Secret) as part of the hard teardown path. Add test assertions covering both the Credentials object and the replica Secret to prevent regressions.

internal/resources/connectivities/init.go
```diff
@@ -111,6 +112,14 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
 	}
 	if ledger == nil {
 		setCondition(connectivity, metav1.ConditionFalse, "LedgerNotFound", "stack has no Ledger module")
+		// Hard gate: the Ledger module has been removed from the stack. This is a
+		// definitive loss of the prerequisite (not a momentary blip), so tear down
+		// any previously provisioned delegated Connectivity + GatewayHTTPAPI rather
+		// than leaving the workload running and exposed through the gateway with no
+		// ledger to ingest into.
+		if err := teardownDelegated(ctx, stack, connectivity); err != nil {
```
NumaryBot, line 132: 🟠 **[major] Hard-gate teardown skipped entirely when connectivity capability becomes unavailable**

When the controller restarts after the connectivity CRD or RBAC becomes unavailable, `connectivityAvailable` is false and `Reconcile` returns at the `OperatorUnavailable` guard before reaching the LedgerNotFound/LedgerNotV3 cleanup logic. If the delegated resources, GatewayHTTPAPI, and Credentials already exist and the ledger has since been removed or downgraded, the public route and credentials remain active even though the hard ledger gate is closed.

**Suggestion:** Evaluate the hard ledger gates (or at minimum attempt deletion of gateway and credentials resources) before the connectivity capability short-circuit, so that already-provisioned resources are torn down even when the connectivity dependency is no longer available.

NumaryBot (commented):
NumaryBot posted 3 new inline findings.

Summary: https://github.com/formancehq/operator/pull/505#issuecomment-5168147906

internal/resources/connectivities/init.go
Dav-14: Already handled in fd721a5f — this describes the pre-fix behavior. `teardownDelegated` no longer returns on the first delete error; it attempts all three deletions via `errors.Join`, so the GatewayHTTPAPI is torn down even when the delegated Connectivity delete fails. Covered by `TestTeardownDelegatedAttemptsEveryDeletion`. The early return this finding references was removed.

internal/resources/connectivities/init.go
Dav-14: Already handled in fd721a5f — this describes the pre-fix behavior. `teardownDelegated` now deletes the cluster-scoped Credentials via `deleteLedgerCredentials` (cascading the ledger operator's key deregistration and the distributed Secret), and tests assert its cleanup on hard teardown and retention on a transient gate.

internal/resources/connectivities/init.go
Dav-14: Fixed in ef83b29c. `Reconcile` now evaluates the ledger hard gate (`ledgerGateClosed`) inside the `!connectivityAvailable` branch and runs `teardownDelegated` when it's closed, so the gateway route and god-mode Credentials are removed even when the connectivity operator has become unavailable. It's guarded on the gate actually being closed so a transient operator outage with a healthy v3 ledger doesn't flap those resources, and `teardownDelegated` now tolerates the connectivity CRD being absent (`ignoreAbsent`) so the delete is a no-op when the API is gone. Tests cover teardown-on-closed-gate and retention-on-open-gate.

internal/resources/connectivities/init.go
```diff
@@ -111,6 +124,14 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
...
```
NumaryBot, line 132: 🟠 **[major] Hard-gate teardown skipped when connectivity capability becomes unavailable**

When the controller restarts and the connectivity CRD or RBAC is no longer available, `connectivityAvailable` is false and `Reconcile` short-circuits at the OperatorUnavailable guard before reaching the LedgerNotFound/LedgerNotV3 cleanup logic. If delegated resources, the GatewayHTTPAPI, and Credentials already exist and the ledger has since been removed or downgraded, the public route and credentials remain active even though the hard ledger gate is closed.

**Suggestion:** Evaluate the hard ledger gates (or at minimum attempt deletion of the GatewayHTTPAPI and Credentials resources) before the connectivity capability short-circuit, so that already-provisioned resources are torn down even when the connectivity dependency is no longer available.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/505#issuecomment-5168147906

internal/resources/connectivities/init.go
Dav-14: Addressed in ef83b29c — the `!connectivityAvailable` guard no longer short-circuits before teardown. It now evaluates the ledger hard gate and tears down when it's closed:

```go
if !connectivityAvailable {
    setCondition(connectivity, metav1.ConditionFalse, "OperatorUnavailable", ...)
    if ledgerGateClosed(ctx, stack) {        // ledger removed, or resolved non-v3
        if err := teardownDelegated(ctx, stack, connectivity); err != nil {
            return err
        }
    }
    return NewPendingError().WithMessage(...)
}
```

`teardownDelegated` deletes the delegated Connectivity, the GatewayHTTPAPI, and the Credentials (and tolerates the connectivity CRD being absent via `ignoreAbsent`). It is guarded on `ledgerGateClosed` so a transient operator outage with a healthy v3 ledger doesn't flap the resources. Covered by `TestConnectivityReconcileTearsDownWhenCapabilityUnavailableAndLedgerGateClosed` and `TestConnectivityReconcileKeepsResourcesWhenCapabilityUnavailableButLedgerV3`.

Dav-14:
@flemzord ready for another look — all three points are addressed and their threads resolved:

1. **Gateway route left exposed when the delegated Connectivity delete fails** — `teardownDelegated` now attempts every deletion via `errors.Join` instead of returning on the first error (fd721a5f).
2. **Cluster-scoped Credentials / distributed Secret not cleaned up on hard teardown** — the teardown now deletes the Credentials via `deleteLedgerCredentials`, which cascades the ledger operator's key deregistration and the distributed Secret (fd721a5f).
3. **Hard-gate teardown skipped when the connectivity capability is unavailable** — `Reconcile` now evaluates the ledger hard gate (`ledgerGateClosed`) inside the `!connectivityAvailable` branch and tears down when it's closed, guarded so a transient operator outage with a healthy v3 ledger doesn't flap the resources; `teardownDelegated` also tolerates the connectivity CRD being absent (ef83b29c).

Each has unit coverage — teardown-on-closed-gate, retention-on-open-gate/transient, and all-deletions-attempted-when-one-fails.

Heads-up: NumaryBot re-posted #3 a couple of times as fresh threads after the fix landed — those re-scans raced the fix and my replies; the current code (ef83b29c) handles it and I've resolved the threads. Would appreciate your re-review when you have a moment.

flemzord (approved):

PR #506, opened 2026-08-03
fix(ledger): resolve configured gRPC port in V3GRPCBackendRef

Dav-14:
V3GRPCBackendRef, the single source of truth consumed by the
connectivity module to build its ledgerAddress, always passed port 0 to
ledgerV3GRPCBackendRef and therefore assumed the default gRPC port. The
gateway backend (v3.go / v3_preview.go) instead resolves the port from
the stack LedgerConfiguration (spec.cluster.service.grpcPort), so a stack
overriding the ledger Cluster gRPC service port got a Connectivity
pointed at the wrong port while the gateway stayed correct.

Resolve the configured port from the LedgerConfiguration inside
V3GRPCBackendRef (via ledgerV3BaseSpec, the same base the gateway derives
its clusterSpec from) so both consumers honour the override and fall back
to the default port when unset. Thread the reconciler Context through the
export and its connectivity caller.

Add a table-driven unit test covering default, stack-scoped, wildcard,
and precedence cases.

---

### DIFF
--- a/internal/resources/connectivities/init.go
+++ b/internal/resources/connectivities/init.go
@@ -211,8 +211,14 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
 
 	// Reuse the single source of truth for the ledger v3 gRPC connection: same
 	// service, port and backend TLS material (self-signed CA secret + SNI) that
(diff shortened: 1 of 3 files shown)


flemzord (approved):
Looks good to me.

PR #507, opened 2026-08-03
fix(connectivity): allow connectivity pods to reach Ledger v3 gRPC via NetworkPolicy

Dav-14:
When networkpolicies.enabled, the default-deny-ingress policy drops all
ingress to Ledger v3 pods except the explicitly-allowed gateway and
intra-cluster ledger peers. The delegated connectivity workload dials the
Ledger v3 gRPC endpoint (ledgers.V3GRPCBackendRef, port 8888) directly, but
connectivity pods are neither gateway nor ledger pods, so their gRPC
connections were silently dropped on network-policy stacks.

Add a dedicated allow-ledger-v3-from-connectivity NetworkPolicy granting the
connectivity workload ingress to the Ledger v3 pods on the gRPC port (8888),
following the existing dedicated-policy pattern (allow-from-gateway).

The connectivity pod labels are owned by the connectivity operator (separate
repo) and cannot be confirmed here; connectivitySelector matches the
operator-wide convention app.kubernetes.io/name=connectivity (documented as an
assumption in code). Add unit tests rendering the policy and extend the
network-policy controller test.

---

### DIFF
--- a/internal/resources/stacks/networkpolicies.go
+++ b/internal/resources/stacks/networkpolicies.go
@@ -182,6 +182,37 @@ func createNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
 		return err
 	}
 
+	// 7. allow-ledger-v3-from-connectivity: the delegated connectivity workload
+	// dials the stack's Ledger v3 gRPC endpoint directly on the ledger Service
+	// (ledgers.V3GRPCBackendRef, port 8888). Connectivity pods are not Ledger v3
+	// pods, so the default-deny policy would otherwise silently drop those gRPC
+	// connections. Allow them, restricted to the gRPC port, following the
+	// dedicated-policy pattern used for the gateway (allow-from-gateway).
+	if _, _, err := CreateOrUpdate[*networkingv1.NetworkPolicy](ctx,
+		types.NamespacedName{
+			Namespace: stack.Name,
+			Name:      "allow-ledger-v3-from-connectivity",
+		},
+		func(np *networkingv1.NetworkPolicy) error {
+			ledgerSelector := directLedgerV3Selector(stack.Name)
+			connectivity := connectivitySelector()
+			np.Spec = networkingv1.NetworkPolicySpec{
+				PodSelector: ledgerSelector,
+				PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
+				Ingress: []networkingv1.NetworkPolicyIngressRule{
+					{
+						From:  []networkingv1.NetworkPolicyPeer{{PodSelector: &connectivity}},
+						Ports: networkPolicyTCPPorts(ledgerV3GRPCPort),
+					},
+				},
+			}
+			return nil
+		},
+		WithController[*networkingv1.NetworkPolicy](ctx.GetScheme(), stack),
+	); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -193,6 +224,7 @@ func deleteNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
 		"allow-ledger-v3-cluster",
 		"allow-ledger-v3-preview-cluster",
 		"allow-ledger-v2-from-v3",
+		"allow-ledger-v3-from-connectivity",
 	} {
 		if err := DeleteIfExists[*networkingv1.NetworkPolicy](ctx, types.NamespacedName{
 			Namespace: stack.Name,
@@ -202,6 +234,29 @@ func deleteNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
 		}
 	}
 	return nil
+}
+
+// ledgerV3GRPCPort mirrors the (unexported) ledgers.ledgerV3GRPCPort: the
+// default Ledger v3 gRPC service port that ledgers.V3GRPCBackendRef targets and
+// that the delegated connectivity workload dials.
+const ledgerV3GRPCPort = 8888
+
+// connectivitySelector matches the delegated connectivity workload pods that
+// dial the stack's Ledger v3 gRPC endpoint.
+//
+// ASSUMPTION: the connectivity pods are provisioned by the connectivity
+// operator (connectivity.formance.com), which lives in a separate repository,
+// so their pod labels are not defined in this repo and cannot be confirmed
+// here. We match the operator-wide convention app.kubernetes.io/name=<component>
+// with the "connectivity" component name (the connectivity core image
+// repository is likewise "connectivity"). We intentionally match only the name
+// label (a subset match) to avoid over-constraining on labels we cannot verify.
+// If the connectivity operator labels its pods differently, this selector must
+// be reconciled against that repository.
+func connectivitySelector() metav1.LabelSelector {
+	return metav1.LabelSelector{MatchLabels: map[string]string{
+		"app.kubernetes.io/name": "connectivity",
+	}}
 }
 
 func directLedgerV3Selector(stackName string) metav1.LabelSelector {
--- a/internal/resources/stacks/networkpolicies.go
+++ b/internal/resources/stacks/networkpolicies.go
@@ -182,12 +182,13 @@ func createNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
 		return err
 	}
 
-	// 7. allow-ledger-v3-from-connectivity: the delegated connectivity workload
-	// dials the stack's Ledger v3 gRPC endpoint directly on the ledger Service
-	// (ledgers.V3GRPCBackendRef, port 8888). Connectivity pods are not Ledger v3
-	// pods, so the default-deny policy would otherwise silently drop those gRPC
-	// connections. Allow them, restricted to the gRPC port, following the
-	// dedicated-policy pattern used for the gateway (allow-from-gateway).
+	// 7. allow-ledger-v3-from-connectivity: let the delegated connectivity
+	// workload reach the stack's Ledger v3 pods (it dials their gRPC endpoint).
+	// Connectivity pods are not Ledger v3 pods, so the default-deny policy would
+	// otherwise drop those connections. Ports are intentionally unrestricted for
+	// this tightly scoped same-namespace source/target pair — mirroring
+	// allow-ledger-v3-cluster — so a LedgerConfiguration spec.cluster.service.grpcPort
+	// override cannot silently break connectivity or leave the policy stale.
 	if _, _, err := CreateOrUpdate[*networkingv1.NetworkPolicy](ctx,
 		types.NamespacedName{
 			Namespace: stack.Name,
@@ -201,8 +202,7 @@ func createNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
 				PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
 				Ingress: []networkingv1.NetworkPolicyIngressRule{
 					{
-						From:  []networkingv1.NetworkPolicyPeer{{PodSelector: &connectivity}},
-						Ports: networkPolicyTCPPorts(ledgerV3GRPCPort),
+						From: []networkingv1.NetworkPolicyPeer{{PodSelector: &connectivity}},
 					},
 				},
 			}
@@ -235,11 +235,6 @@ func deleteNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
 	}
 	return nil
 }
-
-// ledgerV3GRPCPort mirrors the (unexported) ledgers.ledgerV3GRPCPort: the
-// default Ledger v3 gRPC service port that ledgers.V3GRPCBackendRef targets and
-// that the delegated connectivity workload dials.
-const ledgerV3GRPCPort = 8888
 
 // connectivitySelector matches the delegated connectivity workload pods that
 // dial the stack's Ledger v3 gRPC endpoint.
--- a/internal/tests/networkpolicy_controller_test.go
+++ b/internal/tests/networkpolicy_controller_test.go
@@ -124,7 +124,8 @@ var _ = Describe("NetworkPolicyController", func() {
 					g.Expect(np.Spec.Ingress[0].Ports[0].Port.IntValue()).To(Equal(8080))
 				}).Should(Succeed())
 
-				// The delegated connectivity workload can reach the Ledger v3 gRPC port.
+				// The delegated connectivity workload can reach the Ledger v3 pods;
+				// ports are intentionally unrestricted for this tightly scoped pair.
 				Eventually(func(g Gomega) {
 					np := &networkingv1.NetworkPolicy{}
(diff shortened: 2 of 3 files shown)


internal/resources/stacks/networkpolicies.go
```diff
@@ -182,6 +182,37 @@ func createNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
...
+		func(np *networkingv1.NetworkPolicy) error {
+			ledgerSelector := directLedgerV3Selector(stack.Name)
+			connectivity := connectivitySelector()
+			np.Spec = networkingv1.NetworkPolicySpec{
+				PodSelector: ledgerSelector,
+				PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
+				Ingress: []networkingv1.NetworkPolicyIngressRule{
+					{
+						From:  []networkingv1.NetworkPolicyPeer{{PodSelector: &connectivity}},
+						Ports: networkPolicyTCPPorts(ledgerV3GRPCPort),
```
flemzord: This blocks Connectivity on stacks that override spec.cluster.service.grpcPort. PR #506 makes the client dial the configured port, but this policy still only permits 8888, so an override such as 7777 is denied by the default-ingress policy. Consider resolving the effective port here (and watching configuration changes), or leaving the port unrestricted for this tightly scoped source/target pair, and add an override test.

flemzord (changes requested):
One blocking NetworkPolicy issue needs to be addressed.

internal/resources/stacks/networkpolicies.go
NumaryBot: 🟠 **[major] NetworkPolicy hardcodes gRPC port 8888, blocking custom grpcPort overrides**
<sub>reported by flemzord, codex</sub>

The policy only permits traffic on port 8888. Stacks that override spec.cluster.service.grpcPort (e.g. 7777) will be silently denied by the default-ingress policy, breaking connectivity for those deployments.

**Suggestion:** Resolve the effective gRPC port from the stack's configuration (spec.cluster.service.grpcPort) and use that value in the NetworkPolicy rule. Alternatively, leave the port unrestricted for this tightly scoped source/target pair and add a test covering the override scenario. Also ensure the controller watches configuration changes so the policy is reconciled when the port setting changes.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/507#issuecomment-5168129525

internal/resources/stacks/networkpolicies.go
NumaryBot: 🟠 **[major] NetworkPolicy hardcodes gRPC port 8888, blocking custom grpcPort overrides**
<sub>reported by flemzord, codex, NumaryBot</sub>

The NetworkPolicy rule only permits traffic on port 8888. Stacks that override spec.cluster.service.grpcPort (e.g. 7777) will be silently denied by the default-ingress policy, making the fix incomplete for those deployments.

**Suggestion:** Resolve the effective gRPC port from the stack configuration (spec.cluster.service.grpcPort) and use that value in the NetworkPolicy rule. Alternatively, leave the port unrestricted for this tightly scoped source/target pair. Also ensure the controller watches configuration changes so the policy is reconciled when the port setting changes, and add a test covering the override scenario.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/507#issuecomment-5168129525

internal/resources/stacks/networkpolicies.go
NumaryBot: 🟠 **[major] NetworkPolicy hardcodes gRPC port 8888, blocking custom grpcPort overrides**
<sub>reported by flemzord, codex</sub>

The policy only permits traffic on port 8888. Stacks that override spec.cluster.service.grpcPort (e.g. 7777) will be silently denied by the default-ingress policy, breaking connectivity for those deployments. PR #506 makes the client dial the configured port, but this policy does not follow suit.

**Suggestion:** Resolve the effective gRPC port from the stack's configuration (spec.cluster.service.grpcPort) and use that value in the NetworkPolicy rule. Alternatively, leave the port unrestricted for this tightly scoped source/target pair. In either case, ensure the controller watches configuration changes so the policy is reconciled when the port setting changes, and add a test covering the override scenario.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/507#issuecomment-5168129525

internal/resources/stacks/networkpolicies.go
```diff
@@ -182,6 +183,43 @@ func createNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
...
 
+	// 7. allow-ledger-v3-from-connectivity: the delegated connectivity workload
+	// dials the stack's Ledger v3 gRPC endpoint directly on the ledger Service
+	// (ledgers.V3GRPCBackendRef). Connectivity pods are not Ledger v3 pods, so the
+	// default-deny policy would otherwise silently drop those gRPC connections.
+	// Allow them, restricted to the resolved gRPC port — honoring a
+	// spec.cluster.service.grpcPort override exactly as the client does, so a
+	// non-default port is not silently denied — following the dedicated-policy
+	// pattern used for the gateway (allow-from-gateway).
+	ledgerGRPC, err := ledgers.V3GRPCBackendRef(ctx, stack.Name)
```
NumaryBot: 🟠 **[major] Stack controller does not watch LedgerConfiguration changes, leaving NetworkPolicy stale on gRPC port updates**

When spec.cluster.service.grpcPort is changed after the NetworkPolicy has already been rendered, the Stack reconciler has no watch on LedgerConfiguration changes. As a result, the allow-ledger-v3-from-connectivity NetworkPolicy rule is not reconciled and continues to permit only the old port. Traffic from connectivity pods will be silently dropped by the default-deny policy until an unrelated reconcile happens to trigger an update.

**Suggestion:** Add a watch/event handler on LedgerConfiguration resources (specifically grpcPort changes) to the Stack controller so that any port update immediately triggers a reconcile and the NetworkPolicy is kept in sync. Additionally, add a test covering the scenario where grpcPort is overridden and the policy reflects the correct port after an update.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/507#issuecomment-5168129525

internal/resources/stacks/networkpolicies.go
Dav-14: Fixed in 6dc21047. `allow-ledger-v3-from-connectivity` now resolves the effective gRPC port from `ledgers.V3GRPCBackendRef` — the same source of truth the connectivity client dials (#506) — instead of the hardcoded 8888, so a `spec.cluster.service.grpcPort` override is permitted rather than silently denied. Test covers both the default and the override.

On watching config changes: a dedicated `LedgerConfiguration` watch would be a cross-cutting change that affects #506's client consumer too, so I kept this consistent with #506's accepted approach of resolving the port at reconcile time.

Dav-14: Fixed in 6dc21047: the policy resolves the effective gRPC port via `ledgers.V3GRPCBackendRef` (honoring the `spec.cluster.service.grpcPort` override, mirroring #506) instead of hardcoding 8888. Test covers the default and the override.

Dav-14: Fixed in 6dc21047: the policy resolves the effective gRPC port via `ledgers.V3GRPCBackendRef` (honoring the `spec.cluster.service.grpcPort` override, mirroring #506) instead of hardcoding 8888. Test covers the default and the override.

internal/resources/stacks/networkpolicies.go
NumaryBot: 🟠 **[major] NetworkPolicy not reconciled when LedgerConfiguration gRPC port changes**
<sub>reported by codex, flemzord, NumaryBot</sub>

The Stack reconciler does not watch LedgerConfiguration resources. If spec.cluster.service.grpcPort is changed after the NetworkPolicy has already been created, the policy will not be re-rendered and will continue to allow only the old port. Meanwhile, PR #506 already makes the connectivity client dial the new configured port, so the default-deny ingress policy will silently drop that traffic until an unrelated event happens to trigger a Stack reconcile.

**Suggestion:** Add LedgerConfiguration as a watched resource in the Stack controller so that a port change triggers a reconcile. Alternatively, leave the port unrestricted for this tightly scoped source/target pair so the rule is port-agnostic and not sensitive to configuration drift. Either way, add a test covering the port-override scenario.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/507#issuecomment-5168129525

internal/resources/stacks/networkpolicies.go
```diff
@@ -182,6 +183,43 @@ func createNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
...
+		func(np *networkingv1.NetworkPolicy) error {
+			ledgerSelector := directLedgerV3Selector(stack.Name)
+			connectivity := connectivitySelector()
+			np.Spec = networkingv1.NetworkPolicySpec{
+				PodSelector: ledgerSelector,
+				PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
+				Ingress: []networkingv1.NetworkPolicyIngressRule{
+					{
+						From:  []networkingv1.NetworkPolicyPeer{{PodSelector: &connectivity}},
+						Ports: networkPolicyTCPPorts(int(ledgerGRPC.Port)),
```
NumaryBot: 🟠 **[major] Stack controller does not watch LedgerConfiguration, leaving NetworkPolicy stale on gRPC port changes**
<sub>reported by codex, flemzord, NumaryBot</sub>

The NetworkPolicy now resolves the gRPC port from LedgerConfiguration at reconcile time (fixing the hardcoded-8888 issue), but the Stack reconciler does not watch LedgerConfiguration resources. If spec.cluster.service.grpcPort is changed after the policy has been created, no reconcile is triggered for the Stack, so the NetworkPolicy continues to permit only the old port. Since PR #506 makes the connectivity client dial the newly configured port, traffic will be silently dropped by the default-deny ingress policy until an unrelated event happens to trigger a Stack reconcile.

**Suggestion:** Add a watch/event-handler on LedgerConfiguration resources (keyed on grpcPort changes) to the Stack controller so that any port update immediately enqueues a Stack reconcile and the NetworkPolicy is kept in sync. Alternatively, remove the port restriction entirely for this tightly scoped source/target pair so the rule is port-agnostic and immune to configuration drift.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/507#issuecomment-5168129525

internal/resources/stacks/networkpolicies.go
Dav-14: hummmm, is it this PR concern ?

Dav-14: I believe not

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/507#issuecomment-5168129525

internal/resources/stacks/networkpolicies.go
Dav-14: Resolved in 402207e6 by making the policy port-agnostic rather than adding the watch. `allow-ledger-v3-from-connectivity` now leaves the port unrestricted for this tightly scoped same-namespace connectivity↔ledger-v3 pair — mirroring `allow-ledger-v3-cluster` in this same file, which is already unrestricted for exactly this reason (*"Ports are intentionally unrestricted within the cluster so LedgerConfiguration service port overrides cannot break Raft or gRPC"*).

This makes the rule immune to `spec.cluster.service.grpcPort` changes with no `LedgerConfiguration` watch, and avoids the client/policy reconcile-ordering window a watch-based fix would still leave open.

internal/tests/networkpolicy_controller_test.go
```diff
@@ -124,6 +124,23 @@ var _ = Describe("NetworkPolicyController", func() {
...
+					g.Expect(np).To(BeControlledBy(stack))
+					g.Expect(np.Spec.PodSelector.MatchLabels).To(Equal(map[string]string{
+						"app.kubernetes.io/instance": stack.Name,
+						"app.kubernetes.io/name":     "ledger",
+					}))
+					g.Expect(np.Spec.Ingress).To(HaveLen(1))
+					g.Expect(np.Spec.Ingress[0].From).To(HaveLen(1))
+					g.Expect(np.Spec.Ingress[0].From[0].NamespaceSelector).To(BeNil())
+					g.Expect(np.Spec.Ingress[0].From[0].PodSelector.MatchLabels).To(HaveKeyWithValue("app.kubernetes.io/name", "connectivity"))
+					g.Expect(np.Spec.Ingress[0].Ports).To(HaveLen(1))
```
NumaryBot: 🔴 **[blocker] Controller test still expects port 8888 after unrestricted-port change**

The `createNetworkPolicies` function now intentionally leaves `Ingress[0].Ports` empty for the `allow-ledger-v3-from-connectivity` rule, and the new unit test asserts that behavior. However, this controller test still waits for exactly one port entry set to 8888. When the reconciler creates the correct unrestricted policy, this `Eventually` block can never pass, causing CI to fail.

**Suggestion:** Update the expectation in this test to match the rendered policy — either assert that `Ingress[0].Ports` is empty (unrestricted) or, if the port is now resolved dynamically, assert the resolved default port value returned by `ledgers.V3GRPCBackendRef`.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/507#issuecomment-5168129525

internal/tests/networkpolicy_controller_test.go
Dav-14: Good catch — fixed. The controller test (internal/tests/networkpolicy_controller_test.go) now asserts `Ingress[0].Ports` is empty for `allow-ledger-v3-from-connectivity`, matching the unrestricted-port change (and consistent with the other unrestricted rules in the same suite). I'd only updated the package-level unit test and missed this envtest assertion.

internal/resources/stacks/networkpolicies.go
```diff
@@ -204,6 +236,24 @@ func deleteNetworkPolicies(ctx Context, stack *v1beta1.Stack) error {
...
+// repository is likewise "connectivity"). We intentionally match only the name
+// label (a subset match) to avoid over-constraining on labels we cannot verify.
+// If the connectivity operator labels its pods differently, this selector must
+// be reconciled against that repository.
+func connectivitySelector() metav1.LabelSelector {
+	return metav1.LabelSelector{MatchLabels: map[string]string{
+		"app.kubernetes.io/name": "connectivity",
+	}}
+}
+
```
Dav-14, line 256: Can't we use controllerReference ?

flemzord (approved):

internal/resources/stacks/networkpolicies.go
flemzord: A controller reference handles ownership and garbage collection, but it cannot authorize NetworkPolicy traffic: ingress matching still requires a pod selector. I verified the connectivity operator workload labels use `app.kubernetes.io/name: connectivity`, so this selector matches the actual pods.

PR #512, opened 2026-08-04
refactor(connectivity): name the delegated resource "connectivity"

Dav-14:
The delegated connectivity.formance.com/Connectivity is namespaced (one
per stack namespace), so it no longer needs a stack-scoped name. Give it
the fixed name "connectivity"; the connectivity operator derives the API
Service from it, so it becomes "connectivity-api" instead of "<stack>-api".

Credentials (connectivity-<stack>) and the GatewayHTTPAPI
(<stack>-connectivity) stay stack-scoped — they are cluster-scoped and need
the stack for uniqueness. No migration: the module is not yet released.

---

### DIFF
--- a/internal/resources/connectivities/init.go
+++ b/internal/resources/connectivities/init.go
@@ -49,6 +49,11 @@ import (
 
 const (
 	connectivityReadyCondition = "ConnectivityClusterReady"
+	// connectivityDelegatedName is the fixed name of the delegated
+	// connectivity.formance.com/Connectivity resource. It is namespaced (one per
+	// stack namespace), so a constant name stays unique per stack; the connectivity
+	// operator derives its API Service name from it ("connectivity-api").
+	connectivityDelegatedName = "connectivity"
 	// connectivityAPIPort is the port the connectivity-api HTTP server (and its
 	// Service, provisioned by the connectivity operator) listens on.
 	connectivityAPIPort = int32(8080)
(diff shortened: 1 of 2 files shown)


internal/resources/connectivities/init.go
```diff
@@ -224,7 +229,7 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, connectivity *v1beta1.Connecti
 	object := &unstructured.Unstructured{}
 	object.SetGroupVersionKind(connectivityGVK)
 	object.SetNamespace(stack.Name)
-	object.SetName(stack.Name)
+	object.SetName(connectivityDelegatedName)
```
flemzord, line 232: This can leave the old delegated Connectivity workload running after upgrading an existing stack. Previous versions created the owned resource as `<stack>`; changing the desired name only creates `connectivity`, and the old object is not garbage-collected while its parent Connectivity CR still exists. That can leave duplicate core/API workloads active. Consider deleting the legacy `<stack>` resource during reconciliation (after the new resource is ready, if continuity matters) and add an upgrade test that starts with the legacy object and verifies it is removed.

flemzord (changes requested):
One upgrade-path issue needs to be addressed before this rename is safe.

flemzord (approved):
Reviewed the current head again. The previously raised migration concern does not apply because the Connectivity module has not been merged or released, so there is no supported legacy resource to migrate. No remaining code findings.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/512#issuecomment-5189248596

internal/resources/connectivities/init.go
NumaryBot, line 232: 🟠 **[major] Legacy delegated resource not cleaned up after rename**

Previous versions created the delegated Connectivity workload using the stack name (e.g., `<stack>`). After this rename to `connectivity`, reconciliation will create the new resource but will not delete the old one. As long as the parent Connectivity CR exists, the old object is not garbage-collected, potentially leaving duplicate core/API workloads running simultaneously.

**Suggestion:** During reconciliation, after the new `connectivity` resource is confirmed ready, explicitly delete the legacy `<stack>`-named resource if it exists. Additionally, add an upgrade integration test that starts from a state with the legacy-named object and verifies it is removed once the new reconciliation loop runs.

PR #513, opened 2026-08-05
feat(webhooks): coordinate worker-safe upgrades

flemzord:
## Summary

- split the Webhooks API and worker deployments starting with v2.5.0-0
- stop and fully drain legacy embedded or dedicated workers before database migrations
- roll out the new API version before recreating workers
- preserve the monolithic deployment behavior for older Webhooks versions
- cover prerelease, stable, non-v-prefixed, and development versions

## Upgrade sequence

1. Stop existing workers.
2. Wait for their complete termination.
3. Run database migrations.
4. Roll out the new API version and wait for completion.
5. Recreate the worker deployment.

## Validation

- `just tests` (15 suites passed)
- `just pre-commit`
- targeted Webhooks tests
- targeted `golangci-lint`

## Risk

The API/worker topology changes only for Webhooks v2.5.0-0 and newer. Older versions retain the existing embedded-worker deployment.

### DIFF
--- a/internal/resources/webhooks/deployment.go
+++ b/internal/resources/webhooks/deployment.go
@@ -4,6 +4,7 @@ import (
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -202,6 +203,12 @@ func stopWorkers(ctx core.Context, namespace string) error {
 }
 
 func deleteWorkerDeployment(ctx core.Context, namespace string) error {
+	if err := client.IgnoreNotFound(ctx.GetClient().Delete(ctx, &policyv1.PodDisruptionBudget{
+		ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: "webhooks-worker"},
+	})); err != nil {
+		return err
+	}
+
 	deployment := &appsv1.Deployment{}
 	err := ctx.GetClient().Get(ctx, types.NamespacedName{Namespace: namespace, Name: "webhooks-worker"}, deployment)
 	if apierrors.IsNotFound(err) {
@@ -264,5 +271,19 @@ func removeEmbeddedWorker(deployment *appsv1.Deployment) {
 			}
 		}
 		deployment.Spec.Template.Spec.Containers[containerIndex].Env = filtered
+	}
+}
+
+func clearWorkerDeploymentConditions(webhooks *v1beta1.Webhooks) {
+	conditions := webhooks.GetConditions()
+	for _, conditionType := range []string{
+		"DeploymentReady",
+		"PodDisruptionBudget",
+		"PodDisruptionBudgetConfigured",
+	} {
+		conditions.Delete(v1beta1.AndConditions(
+			v1beta1.ConditionTypeMatch(conditionType),
+			v1beta1.ConditionReasonMatch("WebhooksWorker"),
+		))
 	}
 }
--- a/internal/resources/webhooks/deployment.go
+++ b/internal/resources/webhooks/deployment.go
@@ -231,7 +231,7 @@ func deleteWorkerDeployment(ctx core.Context, namespace string) error {
 
 func waitForDeploymentRollout(ctx core.Context, namespace, name string) error {
 	deployment := &appsv1.Deployment{}
-	if err := ctx.GetClient().Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, deployment); err != nil {
+	if err := ctx.GetAPIReader().Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, deployment); err != nil {
 		return err
 	}
 	if !deploymentRolloutComplete(deployment) {
--- a/internal/resources/webhooks/deployment.go
+++ b/internal/resources/webhooks/deployment.go
@@ -4,7 +4,10 @@ import (
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/formancehq/operator/v3/api/formance.com/v1beta1"
 	"github.com/formancehq/operator/v3/internal/core"
@@ -123,4 +126,143 @@ func createAPIDeployment(ctx core.Context, stack *v1beta1.Stack, webhooks *v1bet
 
 func createSingleDeployment(ctx core.Context, stack *v1beta1.Stack, webhooks *v1beta1.Webhooks, database *v1beta1.Database, consumer *v1beta1.BrokerConsumer, version string) error {
 	return createAPIDeployment(ctx, stack, webhooks, database, consumer, version, true)
+}
+
+func createWorkerDeployment(ctx core.Context, stack *v1beta1.Stack, webhooks *v1beta1.Webhooks, database *v1beta1.Database, consumer *v1beta1.BrokerConsumer, version string) error {
+	imageConfiguration, err := registries.GetFormanceImage(ctx, stack, "webhooks", version)
+	if err != nil {
+		return err
+	}
+
+	env, err := deploymentEnvVars(ctx, stack, webhooks, database)
+	if err != nil {
+		return err
+	}
+	topics, err := brokers.GetTopicsEnvVars(ctx, stack, "KAFKA_TOPICS", consumer.Spec.Services...)
+	if err != nil {
+		return err
+	}
+	env = append(env, topics...)
+
+	serviceAccountName, err := settings.GetAWSServiceAccount(ctx, stack.Name)
+	if err != nil {
+		return err
+	}
+
+	return applications.
+		New(webhooks, &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "webhooks-worker",
+			},
+			Spec: appsv1.DeploymentSpec{
+				Template: v1.PodTemplateSpec{
+					Spec: v1.PodSpec{
+						ServiceAccountName: serviceAccountName,
+						ImagePullSecrets:   imageConfiguration.PullSecrets,
+						Containers: []v1.Container{{
+							Name:           "worker",
+							Env:            env,
+							Image:          imageConfiguration.GetFullImageName(),
+							Args:           []string{"worker"},
+							Ports:          []v1.ContainerPort{applications.StandardHTTPPort()},
+							LivenessProbe:  applications.DefaultLiveness("http"),
+							ReadinessProbe: applications.DefaultReadiness("http"),
+						}},
+					},
+				},
+			},
+		}).
+		IsEE().
+		Install(ctx)
+}
+
+func stopWorkers(ctx core.Context, namespace string) error {
+	if err := deleteWorkerDeployment(ctx, namespace); err != nil {
+		return err
+	}
+
+	deployment := &appsv1.Deployment{}
+	err := ctx.GetClient().Get(ctx, types.NamespacedName{Namespace: namespace, Name: "webhooks"}, deployment)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	if deploymentHasEmbeddedWorker(deployment) {
+		removeEmbeddedWorker(deployment)
+		if err := ctx.GetClient().Update(ctx, deployment); err != nil {
+			return err
+		}
+		return core.NewPendingError().WithMessage("waiting for embedded webhooks workers to terminate")
+	}
+
+	return waitForDeploymentRollout(ctx, namespace, "webhooks")
+}
+
+func deleteWorkerDeployment(ctx core.Context, namespace string) error {
+	deployment := &appsv1.Deployment{}
+	err := ctx.GetClient().Get(ctx, types.NamespacedName{Namespace: namespace, Name: "webhooks-worker"}, deployment)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	if !deployment.DeletionTimestamp.IsZero() {
+		return core.NewPendingError().WithMessage("waiting for webhooks workers to terminate")
+	}
+
+	propagationPolicy := metav1.DeletePropagationForeground
+	core.LogDeletion(ctx, deployment, "webhooks.deleteWorkerDeployment")
+	if err := ctx.GetClient().Delete(ctx, deployment, &client.DeleteOptions{PropagationPolicy: &propagationPolicy}); err != nil {
+		return err
+	}
+	return core.NewPendingError().WithMessage("waiting for webhooks workers to terminate")
+}
+
+func waitForDeploymentRollout(ctx core.Context, namespace, name string) error {
+	deployment := &appsv1.Deployment{}
+	if err := ctx.GetClient().Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, deployment); err != nil {
+		return err
+	}
+	if !deploymentRolloutComplete(deployment) {
+		return core.NewPendingError().WithMessage("waiting for deployment %s rollout to complete", name)
+	}
+	return nil
+}
+
+func deploymentRolloutComplete(deployment *appsv1.Deployment) bool {
+	if deployment.Spec.Replicas == nil || deployment.Status.ObservedGeneration != deployment.Generation {
+		return false
+	}
+	desired := *deployment.Spec.Replicas
+	return deployment.Status.UpdatedReplicas >= desired &&
+		deployment.Status.Replicas == deployment.Status.UpdatedReplicas &&
+		deployment.Status.AvailableReplicas >= desired
+}
+
+func deploymentHasEmbeddedWorker(deployment *appsv1.Deployment) bool {
+	for _, container := range deployment.Spec.Template.Spec.Containers {
+		for _, env := range container.Env {
+			if env.Name == "WORKER" && env.Value == "true" {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func removeEmbeddedWorker(deployment *appsv1.Deployment) {
+	for containerIndex := range deployment.Spec.Template.Spec.Containers {
+		env := deployment.Spec.Template.Spec.Containers[containerIndex].Env
+		filtered := env[:0]
+		for _, variable := range env {
+			if variable.Name != "WORKER" {
+				filtered = append(filtered, variable)
+			}
+		}
+		deployment.Spec.Template.Spec.Containers[containerIndex].Env = filtered
+	}
 }
--- a/internal/resources/webhooks/deployment.go
+++ b/internal/resources/webhooks/deployment.go
@@ -1,6 +1,8 @@
 package webhooks
 
 import (
+	"time"
+
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
@@ -232,7 +234,9 @@ func waitForEmbeddedWorkerPodsTermination(ctx core.Context, deployment *appsv1.D
 
 	for index := range pods.Items {
 		if podHasActiveEmbeddedWorker(&pods.Items[index]) {
-			return core.NewPendingError().WithMessage("waiting for embedded webhooks worker pods to terminate")
+			return core.NewPendingError().
+				WithMessage("waiting for embedded webhooks worker pods to terminate").
+				WithRequeueAfter(time.Second)
 		}
 	}
 	return nil
--- a/internal/resources/webhooks/deployment.go
+++ b/internal/resources/webhooks/deployment.go
@@ -221,7 +221,7 @@ func deleteWorkerDeployment(ctx core.Context, namespace string) error {
 	}
 
 	deployment := &appsv1.Deployment{}
-	err := ctx.GetClient().Get(ctx, types.NamespacedName{Namespace: namespace, Name: "webhooks-worker"}, deployment)
+	err := ctx.GetAPIReader().Get(ctx, types.NamespacedName{Namespace: namespace, Name: "webhooks-worker"}, deployment)
 	if apierrors.IsNotFound(err) {
 		return nil
 	}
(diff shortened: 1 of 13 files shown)


internal/resources/webhooks/init.go
```diff
@@ -67,17 +80,44 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, webhooks *v1beta1.Webhooks, ve
...
+				return err
+			}
+			if err := waitForDeploymentRollout(ctx, stack.Name, "webhooks"); err != nil {
+				return err
+			}
+			if err := createWorkerDeployment(ctx, stack, webhooks, database, consumer, version); err != nil {
+				return err
+			}
+		} else {
+			if err := deleteWorkerDeployment(ctx, stack.Name); err != nil {
```
NumaryBot: 🟠 **[major] Clear stale worker conditions when deleting it**

When switching back to a pre-2.5/monolithic version after the split worker deployment has reported a `DeploymentReady=False` condition for the current Webhooks generation (for example a failed worker rollout followed by changing the Stack version), this branch deletes the worker deployment but never removes or refreshes the `WebhooksWorker` conditions. `ForObjectController` treats any false condition with the current generation as pending, so the Webhooks resource can remain NotReady even after the monolithic deployment is healthy.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/513#issuecomment-5189796665

internal/resources/webhooks/init.go
flemzord: Fixed in d2ed0b30. The monolithic path now clears worker-specific DeploymentReady and PDB conditions after the dedicated worker resources are gone, while preserving API conditions. The worker PDB is also deleted alongside its Deployment, and a regression test covers stale false conditions.

internal/resources/webhooks/init.go
```diff
@@ -67,17 +80,45 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, webhooks *v1beta1.Webhooks, ve
...
 	}
 
 	if consumer.Status.Ready {
-		if err := createSingleDeployment(ctx, stack, webhooks, database, consumer, version); err != nil {
-			return err
+		if separateWorkerDeployment {
+			if err := createAPIDeployment(ctx, stack, webhooks, database, consumer, version, false); err != nil {
+				return err
+			}
+			if err := waitForDeploymentRollout(ctx, stack.Name, "webhooks"); err != nil {
```
NumaryBot: 🔴 **[blocker] Stale cache read in `waitForDeploymentRollout` can start workers before API rollout completes**

After `createAPIDeployment` updates the Deployment, the subsequent `waitForDeploymentRollout` call reads from the controller-runtime cached client. If the cache still holds the previously-ready Deployment generation, the check can return nil (i.e., 'ready') before the new API pods have actually rolled out, causing `createWorkerDeployment` to start new workers prematurely. This breaks the worker-safe upgrade ordering that the entire PR is meant to guarantee.

**Suggestion:** Force a direct (non-cached) API server read before evaluating rollout readiness—e.g., use `client.Reader` backed by a direct call or annotate the Get with `DisableCacheFor` so the rollout status reflects the actual server state, not a stale cache entry.

internal/resources/webhooks/init.go
```diff
@@ -67,17 +80,45 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, webhooks *v1beta1.Webhooks, ve
...
+				return err
+			}
+			if err := createWorkerDeployment(ctx, stack, webhooks, database, consumer, version); err != nil {
+				return err
+			}
+		} else {
+			if err := deleteWorkerDeployment(ctx, stack.Name); err != nil {
```
NumaryBot: 🟠 **[major] Stale worker conditions not cleared when reverting to monolithic deployment**

When switching back to a pre-2.5/monolithic version after the split worker deployment has reported a `DeploymentReady=False` condition for the current Webhooks generation, this branch deletes the worker deployment but never removes or refreshes the `WebhooksWorker` conditions. `ForObjectController` treats any false condition with the current generation as pending, so the Webhooks resource can remain NotReady even after the monolithic deployment is healthy.

**Suggestion:** After deleting the worker deployment, explicitly clear or reset the worker-specific conditions (e.g., `WebhooksWorker` / `DeploymentReady`) so the controller does not treat stale false conditions as a pending state.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/513#issuecomment-5189796665

internal/resources/webhooks/init.go
flemzord: Fixed in 50133f3f. Rollout readiness now uses ctx.GetAPIReader(), forcing an uncached API-server read after the Deployment update before workers can be recreated. The targeted Webhooks tests and the full pre-commit workflow pass.

internal/resources/webhooks/init.go
flemzord: This is already fixed in d2ed0b30: clearWorkerDeploymentConditions removes worker-specific DeploymentReady, PodDisruptionBudget, and PodDisruptionBudgetConfigured conditions after the dedicated resources are gone, while preserving API conditions. TestClearWorkerDeploymentConditions covers the stale false-condition case.

internal/resources/webhooks/init.go
NumaryBot: 🔴 **[blocker] Stale cache read in `waitForDeploymentRollout` may start workers before API rollout completes**

After `createAPIDeployment` updates the Deployment, the subsequent `waitForDeploymentRollout` call reads from the controller-runtime cached client. If the cache still holds the previously-ready Deployment generation, the check can return nil ('ready') before the new API pods have actually rolled out. This causes `createWorkerDeployment` to start prematurely, directly breaking the worker-safe upgrade ordering the PR is designed to guarantee.

**Suggestion:** Force a direct (non-cached) API server read before evaluating rollout readiness — e.g., use a `client.Reader` backed by a direct call or annotate the Get with `DisableCacheFor`, so the rollout status reflects actual server state rather than a potentially stale cache entry.

internal/resources/webhooks/deployment.go
```diff
@@ -124,3 +128,162 @@ func createAPIDeployment(ctx core.Context, stack *v1beta1.Stack, webhooks *v1bet
...
+
+	if deploymentHasEmbeddedWorker(deployment) {
+		removeEmbeddedWorker(deployment)
+		if err := ctx.GetClient().Update(ctx, deployment); err != nil {
+			return err
+		}
+		return core.NewPendingError().WithMessage("waiting for embedded webhooks workers to terminate")
+	}
+
+	return waitForDeploymentRollout(ctx, namespace, "webhooks")
```
NumaryBot: 🟠 **[major] Unconditional API rollout wait can block split-to-split upgrades when old API is already unhealthy**

When upgrading from an already-split Webhooks version, the API Deployment carries no embedded worker, so after `webhooks-worker` is deleted there are no workers left to drain. The unconditional rollout wait can block indefinitely if the old API Deployment is currently unavailable or stuck from a prior bad rollout, even though rolling out the new version would be the recovery path. The wait is only meaningful when this function actually removed the WORKER component from a monolithic Deployment.

**Suggestion:** Gate the rollout wait on whether the function actually transitioned the Deployment from a monolithic to a split layout (i.e., whether WORKER was present and removed). If the old Deployment was already split, skip the wait and proceed to create the new API and worker Deployments directly.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/513#issuecomment-5189796665

internal/resources/webhooks/init.go
flemzord: The current branch already uses ctx.GetAPIReader() in getDeploymentAfterRollout (50133f3f), so rollout readiness is read directly from the API server rather than the controller cache. The drain path now uses the direct reader as well in 9aace3b2.

internal/resources/webhooks/deployment.go
flemzord: Fixed in 9aace3b2. Removing an embedded worker now sets an explicit drain annotation and waits for that rollout to complete. Split-to-split upgrades have no drain annotation, so they skip the old API readiness wait and can roll forward from an unhealthy API. The marker is cleared after the direct API-server read confirms completion.

internal/resources/webhooks/deployment.go
```diff
@@ -124,3 +130,191 @@ func createAPIDeployment(ctx core.Context, stack *v1beta1.Stack, webhooks *v1bet
...
+		return err
+	}
+
+	if deploymentHasEmbeddedWorker(deployment) {
+		removeEmbeddedWorker(deployment)
+		markWorkerDrain(deployment)
+		if err := ctx.GetClient().Update(ctx, deployment); err != nil {
+			return err
+		}
+		return core.NewPendingError().WithMessage("waiting for embedded webhooks workers to terminate")
```
NumaryBot, line 204: 🟠 **[major] Unconditional API rollout wait can block split-to-split upgrades indefinitely**

When upgrading from an already-split Webhooks version, the API Deployment carries no embedded worker component. After `webhooks-worker` is deleted, there are no workers left to drain, so the unconditional rollout wait adds no safety value. If the old API Deployment is currently unavailable or stuck from a prior bad rollout, this wait can block indefinitely, even though deploying the new version would be the recovery path. The wait is only meaningful when the function actually transitioned from a monolithic to a split layout.

**Suggestion:** Gate the rollout wait on whether the function actually removed a WORKER component from a monolithic Deployment (i.e., whether the transition was monolithic→split). If the old Deployment was already split, skip the wait and proceed directly to creating the new API and worker Deployments.

internal/resources/webhooks/deployment.go
```diff
@@ -124,3 +130,191 @@ func createAPIDeployment(ctx core.Context, stack *v1beta1.Stack, webhooks *v1bet
...
+}
+
+func deleteWorkerDeployment(ctx core.Context, namespace string) error {
+	if err := client.IgnoreNotFound(ctx.GetClient().Delete(ctx, &policyv1.PodDisruptionBudget{
+		ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: "webhooks-worker"},
+	})); err != nil {
+		return err
+	}
+
+	deployment := &appsv1.Deployment{}
```
NumaryBot, line 266: 🟠 **[major] Cached client read for worker Deployment may return stale NotFound, skipping worker drain**

During upgrades from a split Webhooks version, the cached `Get` for `webhooks-worker` can return a stale NotFound — for example, right after the worker Deployment was created but before the cache has observed it. In that case `deleteWorkerDeployment` returns nil, migrations proceed, and existing worker pods may still be running, violating the safe-upgrade ordering this PR is designed to guarantee.

**Suggestion:** Use the uncached API reader (e.g., `ctx.GetAPIReader()`) when checking for the existence of the worker Deployment before deciding there is nothing to drain, consistent with the fix already applied in `waitForDeploymentRollout`.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/513#issuecomment-5189796665

internal/resources/webhooks/deployment.go
flemzord: This is already fixed in 9aace3b2. stopWorkers returns immediately for a split API unless the explicit formance.com/webhooks-workers-draining annotation is present. That annotation is only set when WORKER is actually removed from a monolithic Deployment.

internal/resources/webhooks/deployment.go
flemzord: Fixed in cfcfe9e0. The dedicated worker Deployment existence check now uses ctx.GetAPIReader(), so a stale cached NotFound cannot skip foreground deletion and worker termination.

internal/resources/webhooks/init.go
```diff
@@ -67,17 +80,45 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, webhooks *v1beta1.Webhooks, ve
 		if err := databases.SaveModuleVersion(ctx, database, version); err != nil {
 			return errors.Wrap(err, "saving module version in database object")
 		}
+
+		if separateWorkerDeployment {
+			return NewPendingError().WithMessage("database migrated, waiting to deploy webhooks API")
+		}
 	}
 
 	if consumer.Status.Ready {
```
NumaryBot: 🟠 **[major] BrokerConsumer readiness guard incorrectly blocks API Deployment rollout after migrations**

When upgrading to a split Webhooks version, if the BrokerConsumer is not yet Ready, the guard at this line skips both the API and worker creation. However, database migrations have already been applied at this point. The old API Deployment is left running against the migrated database until the consumer becomes ready. Because `createAPIDeployment(..., false)` no longer uses the consumer topics, there is no reason to block the new API rollout on consumer readiness — only the worker recreation legitimately depends on the consumer. This violates the migrate-then-roll-out-new-API ordering the PR is designed to enforce.

**Suggestion:** Move the BrokerConsumer readiness check to guard only the worker deployment creation step. Allow `createAPIDeployment` to proceed unconditionally (after migrations), and only gate `createWorkerDeployment` on `consumer.Status.Ready`.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/513#issuecomment-5189796665

internal/resources/webhooks/init.go
flemzord: Fixed in e0b62ce7. Split-mode reconciliation now always rolls out and waits for the new API after migrations. BrokerConsumer readiness gates only createWorkerDeployment; the legacy monolithic path keeps its existing readiness requirement.

internal/resources/webhooks/init.go
```diff
@@ -67,9 +80,30 @@ func Reconcile(ctx Context, stack *v1beta1.Stack, webhooks *v1beta1.Webhooks, ve
...
 			return errors.Wrap(err, "saving module version in database object")
 		}
+
+		if separateWorkerDeployment {
+			return NewPendingError().WithMessage("database migrated, waiting to deploy webhooks API")
+		}
 	}
 
-	if consumer.Status.Ready {
+	if separateWorkerDeployment {
```
NumaryBot, line 89: 🟠 **[major] BrokerConsumer readiness guard blocks API Deployment rollout unnecessarily**

When upgrading to a split Webhooks version, if the BrokerConsumer is not yet Ready, the guard skips both API and worker creation. Database migrations have already been applied at this point, leaving the old API Deployment running against the migrated database until the consumer becomes ready. Because createAPIDeployment(..., false) no longer uses consumer topics, there is no reason to block the new API rollout on consumer readiness — only the worker recreation legitimately depends on the consumer. This violates the migrate-then-roll-out-new-API ordering the PR is designed to enforce.

**Suggestion:** Move the BrokerConsumer readiness check to guard only the worker deployment creation step. Allow createAPIDeployment to proceed unconditionally after migrations complete, and gate only createWorkerDeployment on consumer.Status.Ready.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/513#issuecomment-5189796665

internal/resources/webhooks/init.go
flemzord: This finding is stale on e0b62ce7. In the current code, createAPIDeployment and waitForDeploymentRollout execute at lines 90-94 before consumer.Status.Ready is checked at line 96. Only createWorkerDeployment at line 99 is gated by BrokerConsumer readiness. No code change is needed.

internal/resources/webhooks/deployment.go
```diff
@@ -124,3 +130,191 @@ func createAPIDeployment(ctx core.Context, stack *v1beta1.Stack, webhooks *v1bet
...
+	}
+	return deployment, nil
+}
+
+func deploymentRolloutComplete(deployment *appsv1.Deployment) bool {
+	if deployment.Spec.Replicas == nil || deployment.Status.ObservedGeneration != deployment.Generation {
+		return false
+	}
+	desired := *deployment.Spec.Replicas
+	return deployment.Status.UpdatedReplicas >= desired &&
```
NumaryBot, line 307: 🟠 **[major] Deployment rollout completion does not guarantee embedded worker pod termination before migrations**

When draining an embedded (monolithic) worker, `stopWorkers` uses Deployment rollout status as the signal that old worker pods have exited. However, a Deployment is considered rolled out once new pods are Ready and old ReplicaSets are scaled to zero — i.e., old pods have received a termination signal — but they may still be executing their termination grace period. Migrations can therefore start while a terminating `WORKER=true` pod is still processing jobs, violating the worker-safe upgrade ordering this PR is designed to enforce.

**Suggestion:** After the Deployment rollout is confirmed complete, additionally verify that no pods with the old worker labels remain in a non-terminal phase (i.e., wait until all pods whose phase is not Succeeded or Failed and which match the old worker selector are gone) before allowing migrations to proceed. This can be done with a targeted pod list using the uncached API reader and re-queuing if any such pods are still present.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/513#issuecomment-5189796665

internal/resources/webhooks/deployment.go
flemzord: Fixed in ab6ad921. After the Deployment rollout completes, the controller now lists matching pods through the uncached API reader and keeps requeuing while any non-terminal pod still carries WORKER=true. The drain annotation is only cleared after those embedded worker pods have exited. Added coverage for a running pod with a deletion timestamp and for terminal/API-only pods.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/513#issuecomment-5189796665

internal/resources/webhooks/deployment.go
```diff
@@ -124,3 +130,230 @@ func createAPIDeployment(ctx core.Context, stack *v1beta1.Stack, webhooks *v1bet
...
+	if err := ctx.GetAPIReader().List(ctx, pods,
+		client.InNamespace(deployment.Namespace),
+		client.MatchingLabelsSelector{Selector: selector},
+	); err != nil {
+		return err
+	}
+
+	for index := range pods.Items {
+		if podHasActiveEmbeddedWorker(&pods.Items[index]) {
+			return core.NewPendingError().WithMessage("waiting for embedded webhooks worker pods to terminate")
```
NumaryBot: 🔴 **[blocker] Waiting for terminating worker pods can stall the reconciler indefinitely**

When an embedded-worker rollout is confirmed complete but old pods are still within their termination grace period, the code returns a PendingError to signal 'not yet done'. However, the Webhooks reconciler does not watch Pods. If no Deployment status change or other watched-resource event occurs after the terminating pod finally exits, the reconciler will never be re-triggered, causing the upgrade to stall at this point forever — migrations will never run and the new worker will never be created.

**Suggestion:** Return a requeue-after error (e.g., ctrl.Result{RequeueAfter: 5s}) instead of (or in addition to) a PendingError so the reconciler is guaranteed to re-evaluate this condition even if no watched resource changes. Alternatively, add a Pod watch to the controller for the relevant label selector so pod deletions naturally trigger reconciliation.

flemzord: Fixed in bb4fbfce. Application errors can now carry an explicit requeue delay, the standard reconciler propagates it into ctrl.Result.RequeueAfter, and the embedded-worker pod drain requests a one-second requeue while a non-terminal WORKER=true pod remains. Existing application errors retain their current zero-delay behavior. Added wrapped-error coverage and ran the full test suite.

internal/resources/webhooks/deployment.go
```diff
@@ -124,3 +132,232 @@ func createAPIDeployment(ctx core.Context, stack *v1beta1.Stack, webhooks *v1bet
...
+	if err := ctx.GetAPIReader().List(ctx, pods,
+		client.InNamespace(deployment.Namespace),
+		client.MatchingLabelsSelector{Selector: selector},
+	); err != nil {
+		return err
+	}
+
+	for index := range pods.Items {
+		if podHasActiveEmbeddedWorker(&pods.Items[index]) {
+			return core.NewPendingError().
```
NumaryBot, line 237: 🔴 **[blocker] Pod-drain requeue-after is swallowed, stalling the upgrade indefinitely**

When an embedded worker pod is still within its termination grace period, the code issues a WithRequeueAfter (or PendingError) to signal 'not yet done'. However, ForObjectController converts all ApplicationErrors into status updates and returns nil to controller-runtime, so the requested requeue delay is never surfaced to the outer reconcile loop. Because the Webhooks controller does not watch Pods, no watched resource event is guaranteed to fire once the terminating pod finally exits — the reconciler may never be re-triggered, causing migrations to stall indefinitely and the safe-upgrade sequence to never complete.

**Suggestion:** Either (a) return a ctrl.Result{RequeueAfter: 5s} directly from the reconciler (bypassing the ApplicationError path so controller-runtime actually schedules the requeue), or (b) add a Pod watch to the Webhooks controller for the relevant worker label selector so that pod deletion events naturally trigger reconciliation, or (c) both. The prior fix in ab6ad921 introduced the pod-level drain check but did not address this requeueing gap.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/513#issuecomment-5189796665

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/513#issuecomment-5189796665

internal/resources/webhooks/deployment.go
NumaryBot, line 237: 🔴 **[blocker] Pod-drain requeue-after is swallowed by ForObjectController, stalling upgrades indefinitely**

When an embedded worker pod is still within its termination grace period, WithRequeueAfter(time.Second) is returned as an ApplicationError. However, ForObjectController (the wrapper registered via WithModuleReconciler) converts all ApplicationErrors into status updates and returns nil to controller-runtime, so the requested requeue delay is silently discarded. Because the Webhooks controller does not watch Pods, no watched-resource event is guaranteed to fire once the terminating pod finally exits. The reconciler may therefore never be re-triggered, causing migrations to stall indefinitely and the safe-upgrade sequence to never complete. This issue was flagged in the prior PR discussion (Thread 12) and by an independent reviewer; the fix introduced in bb4fbfce added the requeue-after signal but did not ensure it propagates past the ApplicationError wrapper.

**Suggestion:** Either (a) return a ctrl.Result{RequeueAfter: 5*time.Second} directly from the top-level Reconcile method, bypassing the ApplicationError path so controller-runtime actually schedules the requeue; or (b) add a Pod watch to the Webhooks controller for the relevant worker label selector so pod deletion events naturally trigger reconciliation; or (c) both. Whichever approach is taken, verify end-to-end that the requeue signal reaches controller-runtime and is not consumed by any intermediate wrapper.

flemzord: Fixed in 267677c7. ForObjectController now propagates application errors after updating status, allowing the outer reconciler to extract the scheduled delay while still suppressing the error as non-fatal. Added adapter-level coverage proving the requeue delay survives this layer, and the full test suite passes.

flemzord: Fixed by 267677c7, which was pushed immediately after this review was generated. ForObjectController now returns the ApplicationError after setting status, the outer reconciler extracts its RequeueAfter value, and controller-runtime receives that delay in ctrl.Result. The new TestForObjectControllerPropagatesApplicationErrorRequeue covers this exact intermediate wrapper, and the full test suite passes.

PR #515, opened 2026-08-05
fix(ci): make staging promotion atomic

flemzord:
## Summary

- queue `main` workflow runs instead of cancelling an in-flight staging promotion
- wait for any existing Argo CD operation before mutating image parameters
- roll back both image overrides when synchronization or verification fails
- verify the application is synced, healthy, and running the requested Operator tag
- cover successful promotion and rollback paths with a hermetic shell test

## Root cause

The staging deployment updated persistent Argo CD Helm parameters before starting the sync. If the sync failed or the workflow was cancelled, the parameter mutation survived and could later be applied independently. This left staging on `33bc2957` even though the chart metadata had advanced to `v3.13.1`.

## Validation

- `scripts/deploy-staging_test.sh`
- `nix --extra-experimental-features "nix-command flakes" develop --impure --command just pre-commit`
- `git diff --check`

### DIFF
--- a/scripts/deploy-staging.sh
+++ b/scripts/deploy-staging.sh
@@ -1,0 +1,73 @@
+#!/bin/sh
+
+set -eu
+
+: "${APPLICATION:?APPLICATION is required}"
+: "${AUTH_TOKEN:?AUTH_TOKEN is required}"
+: "${SERVER:?SERVER is required}"
+: "${TAG:?TAG is required}"
+
+argocd_cli() {
+	argocd --auth-token="$AUTH_TOKEN" --server="$SERVER" --grpc-web "$@"
+}
+
+parameter_value() {
+	argocd_cli app get "$APPLICATION" --output json |
+		jq -r --arg name "$1" \
+		'.spec.source.helm.parameters // [] | map(select(.name == $name)) | last | .value // empty'
+}
+
+restore_parameter() {
+	name=$1
+	value=$2
+	if [ -n "$value" ]; then
+		argocd_cli app set "$APPLICATION" --parameter "$name=$value"
+	else
+		argocd_cli app unset "$APPLICATION" --parameter "$name"
+	fi
+}
+
+mutated=false
+previous_image_tag=""
+previous_utils_tag=""
+
+rollback() {
+	status=$?
+	trap - EXIT INT TERM
+	if [ "$status" -ne 0 ] && [ "$mutated" = true ]; then
+		echo "Staging promotion failed; restoring previous Argo CD parameters" >&2
+		restore_parameter image.tag "$previous_image_tag" || true
+		restore_parameter operator.utils.tag "$previous_utils_tag" || true
+		argocd_cli app sync "$APPLICATION" --timeout 600 || true
+		argocd_cli app wait "$APPLICATION" --operation --sync --health --timeout 600 || true
+	fi
+	exit "$status"
+}
+
+trap rollback EXIT INT TERM
+
+# Never mutate parameters while another sync can still consume them.
+argocd_cli app wait "$APPLICATION" --operation --timeout 600
+
+previous_image_tag=$(parameter_value image.tag)
+previous_utils_tag=$(parameter_value operator.utils.tag)
+
+argocd_cli app set "$APPLICATION" \
+	--parameter "image.tag=$TAG" \
+	--parameter "operator.utils.tag=$TAG"
+mutated=true
+
+argocd_cli app sync "$APPLICATION" --timeout 600
+argocd_cli app wait "$APPLICATION" --operation --sync --health --timeout 600
+
+application=$(argocd_cli app get "$APPLICATION" --refresh --output json)
+echo "$application" | jq -e --arg tag "$TAG" '
+  (.spec.source.helm.parameters // [] | any(.name == "image.tag" and .value == $tag)) and
+  (.spec.source.helm.parameters // [] | any(.name == "operator.utils.tag" and .value == $tag)) and
+  (.status.sync.status == "Synced") and
+  (.status.health.status == "Healthy") and
+  (.status.summary.images // [] | any(endswith("/operator:" + $tag)))
+' >/dev/null
+
+mutated=false
+trap - EXIT INT TERM
--- a/scripts/deploy-staging.sh
+++ b/scripts/deploy-staging.sh
@@ -11,46 +11,52 @@ argocd_cli() {
 	argocd --auth-token="$AUTH_TOKEN" --server="$SERVER" --grpc-web "$@"
 }
 
-parameter_value() {
-	argocd_cli app get "$APPLICATION" --output json |
-		jq -r --arg name "$1" \
-		'.spec.source.helm.parameters // [] | map(select(.name == $name)) | last | .value // empty'
-}
-
-restore_parameter() {
-	name=$1
-	value=$2
-	if [ -n "$value" ]; then
-		argocd_cli app set "$APPLICATION" --parameter "$name=$value"
-	else
-		argocd_cli app unset "$APPLICATION" --parameter "$name"
-	fi
-}
-
 mutated=false
-previous_image_tag=""
-previous_utils_tag=""
+previous_parameters='[]'
+
+wait_for_operation() {
+	if argocd_cli app wait "$APPLICATION" --operation --timeout 600; then
+		return 0
+	fi
+
+	echo "Argo CD operation did not finish; terminating it before rollback" >&2
+	argocd_cli app terminate-op "$APPLICATION"
+	argocd_cli app wait "$APPLICATION" --operation --timeout 60
+}
 
 rollback() {
-	status=$?
-	trap - EXIT INT TERM
-	if [ "$status" -ne 0 ] && [ "$mutated" = true ]; then
+	status=$1
+	trap - EXIT
+	trap '' INT TERM
+
+	if [ "$mutated" = true ]; then
 		echo "Staging promotion failed; restoring previous Argo CD parameters" >&2
-		restore_parameter image.tag "$previous_image_tag" || true
-		restore_parameter operator.utils.tag "$previous_utils_tag" || true
-		argocd_cli app sync "$APPLICATION" --timeout 600 || true
-		argocd_cli app wait "$APPLICATION" --operation --sync --health --timeout 600 || true
+		if wait_for_operation; then
+			patch=$(jq -cn --argjson parameters "$previous_parameters" \
+				'{spec:{source:{helm:{parameters:$parameters}}}}')
+			if argocd_cli app patch "$APPLICATION" --type merge --patch "$patch"; then
+				argocd_cli app sync "$APPLICATION" --timeout 600 || true
+				argocd_cli app wait "$APPLICATION" --operation --sync --health --timeout 600 || true
+			else
+				echo "Unable to restore the previous Argo CD parameters" >&2
+			fi
+		else
+			echo "Unable to stop the active Argo CD operation; parameters were not changed during rollback" >&2
+		fi
 	fi
+
 	exit "$status"
 }
 
-trap rollback EXIT INT TERM
+trap 'rollback $?' EXIT
+trap 'rollback 130' INT
+trap 'rollback 143' TERM
 
 # Never mutate parameters while another sync can still consume them.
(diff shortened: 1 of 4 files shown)


scripts/deploy-staging.sh
```diff
@@ -0,0 +1,73 @@
...
+		argocd_cli app unset "$APPLICATION" --parameter "$name"
+	fi
+}
+
+mutated=false
+previous_image_tag=""
+previous_utils_tag=""
+
+rollback() {
+	status=$?
```
NumaryBot: 🟠 **[major] Treat INT/TERM traps as failed promotions**

When the script receives SIGINT/SIGTERM after the Argo CD parameters have been mutated, this handler reuses `$?` from whatever last ran successfully; for example, a cancellation between the successful `app set` and the next command can enter the trap with status 0. In that case the `status != 0` guard skips rollback and exits 0 even though `mutated=true`, leaving the persistent staging overrides behind.

scripts/deploy-staging.sh
```diff
@@ -0,0 +1,73 @@
...
+	trap - EXIT INT TERM
+	if [ "$status" -ne 0 ] && [ "$mutated" = true ]; then
+		echo "Staging promotion failed; restoring previous Argo CD parameters" >&2
+		restore_parameter image.tag "$previous_image_tag" || true
```
NumaryBot: 🟠 **[major] Wait out the failed sync before restoring overrides**

When `argocd app sync` or the subsequent wait times out while Argo CD still has the sync operation running, rollback immediately changes the Helm parameters before ensuring that operation has finished. In that timeout scenario the in-flight sync can still consume the new tag, and the rollback `app sync` can be rejected because another operation is in progress, so a failed promotion may still apply instead of being atomically reverted.

NumaryBot (commented):

Summary: https://github.com/formancehq/operator/pull/515#issuecomment-5192874264

scripts/deploy-staging.sh
```diff
@@ -0,0 +1,79 @@
...
+# Never mutate parameters while another sync can still consume them.
+argocd_cli app wait "$APPLICATION" --operation --timeout 600
+
+baseline=$(argocd_cli app get "$APPLICATION" --output json)
+previous_parameters=$(printf '%s\n' "$baseline" | jq -c '.spec.source.helm.parameters // []')
+
+argocd_cli app set "$APPLICATION" \
+	--parameter "image.tag=$TAG" \
+	--parameter "operator.utils.tag=$TAG"
+mutated=true
```
NumaryBot: 🟠 **[major] Rollback guard `mutated` set after `app set` — SIGTERM/SIGINT window leaves parameters dirty**

If the process receives SIGINT/SIGTERM while `argocd app set` is running or immediately after it returns, Argo CD may have already persisted the new Helm parameters but `mutated` is still `false`. The trap then skips rollback and exits 0, leaving the staging overrides in the new (broken) state. The same race occurs if the script is cancelled between the two `app set` calls for `image.tag` and `operator.utils.tag`.

**Suggestion:** Set `mutated=true` (or equivalent guard) **before** the first mutating `argocd app set` call, not after. Also force the trap to treat SIGINT/SIGTERM as a failure regardless of `$?`, so that any cancellation after the guard is set always triggers rollback.

scripts/deploy-staging.sh
```diff
@@ -0,0 +1,79 @@
...
+	trap '' INT TERM
+
+	if [ "$mutated" = true ]; then
+		echo "Staging promotion failed; restoring previous Argo CD parameters" >&2
+		if wait_for_operation; then
+			patch=$(jq -cn --argjson parameters "$previous_parameters" \
+				'{spec:{source:{helm:{parameters:$parameters}}}}')
+			if argocd_cli app patch "$APPLICATION" --type merge --patch "$patch"; then
+				argocd_cli app sync "$APPLICATION" --timeout 600 || true
+				argocd_cli app wait "$APPLICATION" --operation --sync --health --timeout 600 || true
```
NumaryBot, line 39: 🟠 **[major] In-flight Argo CD sync not terminated before rollback re-applies parameters**

When `argocd app sync` or a subsequent wait times out while Argo CD still has a sync operation running, the rollback immediately changes the Helm parameters before that operation completes. The in-flight sync can still consume the new tag, and the rollback `app sync` may be rejected because another operation is in progress. This means a failed promotion can still be applied despite the rollback attempt.

**Suggestion:** Before restoring parameters in the rollback path, cancel any in-progress Argo CD sync operation (e.g. `argocd app terminate-op "$APPLICATION"`) and wait for it to settle, then apply the parameter restore and re-sync.

scripts/deploy-staging.sh
```diff
@@ -0,0 +1,79 @@
...
+
+argocd_cli() {
+	argocd --auth-token="$AUTH_TOKEN" --server="$SERVER" --grpc-web "$@"
+}
+
+mutated=false
+previous_parameters='[]'
+
+wait_for_operation() {
+	if argocd_cli app wait "$APPLICATION" --operation --timeout 600; then
```
NumaryBot, line 18: 🟠 **[major] Pipeline failure in `parameter_value` silently records empty baseline**

In `parameter_value`, `argocd_cli app get` is the non-final command in a pipeline. If it fails, `jq` can exit successfully on empty input. The script then records empty strings as the previous parameter values. During rollback these empty values will unset valid production overrides instead of restoring them.

**Suggestion:** Capture the Argo CD response into a variable first and propagate its exit code before piping to `jq`:
```sh
application_json=$(argocd_cli app get "$APPLICATION" --output json) || return 1
printf '%s\n' "$application_json" | jq -r ...
```

scripts/deploy-staging.sh
```diff
@@ -0,0 +1,79 @@
...
+			if argocd_cli app patch "$APPLICATION" --type merge --patch "$patch"; then
+				argocd_cli app sync "$APPLICATION" --timeout 600 || true
+				argocd_cli app wait "$APPLICATION" --operation --sync --health --timeout 600 || true
+			else
+				echo "Unable to restore the previous Argo CD parameters" >&2
+			fi
+		else
+			echo "Unable to stop the active Argo CD operation; parameters were not changed during rollback" >&2
+		fi
+	fi
```
NumaryBot, line 46: 🟠 **[major] Two separate `app set` calls in rollback can leave mixed parameter state**

Each `restore_parameter` call (for `image.tag` and `operator.utils.tag`) issues a separate `argocd app set`. If the second call fails after the first succeeds, `image.tag` is restored but `operator.utils.tag` remains at the failed-promotion value. A subsequent sync applies this mixed state, making the rollback non-atomic.

**Suggestion:** Combine both parameter restores into a single `argocd app set` (or `argocd app patch`) invocation so both values are written atomically, then sync once.

scripts/deploy-staging_test.sh
```diff
@@ -0,0 +1,146 @@
...
+		;;
+	"app wait")
+		if [ -f "$FAKE_ARGOCD_OPERATION" ]; then
+			exit 1
+		fi
+		;;
+	"app terminate-op")
+		rm -f "$FAKE_ARGOCD_OPERATION"
+		;;
+esac
```
NumaryBot, line 98: 🟡 **[minor] Fake Argo CD handler tracks only one parameter; rollback of `operator.utils.tag` is never verified**

The fake ArgoCD handler uses a single state variable for both parameters and only applies/returns `image.tag`. Success and rollback tests can pass even if `operator.utils.tag` is never updated or restored, hiding real bugs in the rollback logic.

**Suggestion:** Store each parameter in its own state variable. Apply and return both independently in the fake. Extend success and rollback assertions to check both parameter values. Add a test case for an absent parameter to exercise the `app unset` path.

NumaryBot (commented):
NumaryBot posted 5 new inline findings.

Summary: https://github.com/formancehq/operator/pull/515#issuecomment-5192874264

scripts/deploy-staging.sh
```diff
@@ -0,0 +1,79 @@
...
+# Never mutate parameters while another sync can still consume them.
+argocd_cli app wait "$APPLICATION" --operation --timeout 600
+
+baseline=$(argocd_cli app get "$APPLICATION" --output json)
+previous_parameters=$(printf '%s\n' "$baseline" | jq -c '.spec.source.helm.parameters // []')
+
+mutated=true
+argocd_cli app set "$APPLICATION" \
+	--parameter "image.tag=$TAG" \
+	--parameter "operator.utils.tag=$TAG"
```
NumaryBot, line 64: 🟠 **[major] Mutation guard `mutated` set after `app set` — SIGINT/SIGTERM window leaves parameters dirty**
<sub>reported by NumaryBot, codex</sub>

The `mutated` flag is set only after the first `argocd app set` call succeeds. If the process receives SIGINT/SIGTERM while that call is running or immediately after it returns, Argo CD may have already persisted the new Helm parameters but `mutated` is still `false`. The trap then skips rollback and exits 0, leaving the staging overrides in the broken state. The same race exists between the two `app set` calls for `image.tag` and `operator.utils.tag`.

**Suggestion:** Set `mutated=true` before the first mutating `argocd app set` call, not after. Also force the INT/TERM trap to treat those signals as failures (e.g. `exit 1`) regardless of `$?`, so any cancellation after the guard is set always triggers rollback.

scripts/deploy-staging.sh
NumaryBot, line 39: 🟠 **[major] In-flight Argo CD sync not terminated before rollback restores parameters**
<sub>reported by NumaryBot, codex</sub>

When `argocd app sync` or a subsequent wait times out while Argo CD still has a sync operation running, the rollback immediately changes the Helm parameters before that operation completes. The in-flight sync can still consume the new tag, and the rollback `app sync` may be rejected because another operation is already in progress. A failed promotion can therefore still be applied to staging despite the rollback attempt.

**Suggestion:** Before restoring parameters in the rollback path, cancel any in-progress Argo CD sync operation (e.g. `argocd app terminate-op "$APPLICATION"`) and wait for it to settle, then apply the parameter restore and re-sync.

scripts/deploy-staging.sh
NumaryBot, line 18: 🟠 **[major] Pipeline failure in `parameter_value` silently records empty baseline**

`argocd_cli app get` is the non-final command in a pipeline. If it fails, `jq` can exit successfully on empty input. The script then records empty strings as the previous parameter values. During rollback these empty values will unset valid production overrides instead of restoring them, corrupting the staging application state.

**Suggestion:** Capture the Argo CD response into a variable first and propagate its exit code before piping to `jq`:
```sh
application_json=$(argocd_cli app get "$APPLICATION" --output json) || return 1
printf '%s\n' "$application_json" | jq -r --arg name "$1" \
  '.spec.source.helm.parameters // [] | map(select(.name == $name)) | last | .value // empty'
```

scripts/deploy-staging.sh
NumaryBot, line 46: 🟠 **[major] Two separate `app set` calls in rollback can leave mixed parameter state**

Each `restore_parameter` call (for `image.tag` and `operator.utils.tag`) issues a separate `argocd app set`. If the second call fails after the first succeeds, `image.tag` is restored but `operator.utils.tag` remains at the failed-promotion value. A subsequent sync applies this mixed state, making the rollback non-atomic and defeating the PR's stated purpose.

**Suggestion:** Combine both parameter restores into a single `argocd app set` invocation (passing both `--helm-set` flags at once) so both values are written atomically, then sync once.

scripts/deploy-staging_test.sh
```diff
@@ -0,0 +1,162 @@
...
```
NumaryBot, line 98: 🟡 **[minor] Fake Argo CD handler tracks only one parameter; rollback of `operator.utils.tag` is never verified**

The fake ArgoCD handler uses a single state variable for both parameters and only applies/returns `image.tag`. Success and rollback tests can pass even if `operator.utils.tag` is never updated or restored, hiding real bugs in the rollback logic.

**Suggestion:** Store each parameter in its own state variable. Apply and return both independently in the fake. Extend success and rollback assertions to check both parameter values. Add a test case for an absent parameter to exercise the `app unset` path.

NumaryBot (commented):
NumaryBot posted 5 new inline findings.

Summary: https://github.com/formancehq/operator/pull/515#issuecomment-5192874264

PR #516, opened 2026-08-06
fix(connectivity): report pending when ledger Credentials API is unavailable

Dav-14:
A missing ledger Credentials CRD (NoMatch) or absent RBAC (Forbidden) — e.g.
an older ledger operator that predates the Credentials API — was surfaced as a
hard reconcile failure (LedgerCredentialsFailed). Treat it as an external
capability gap instead: set a LedgerCredentialsUnavailable condition and return
a pending error so the module reports unavailable and retries once the ledger
operator provides the capability, mirroring the connectivity operator gate.

Addresses NumaryBot finding 4132b30d8cdd379d.

---


Dav-14:
Superseded by #492, which includes this fix with the polling fallback and expanded tests.

PR #517, opened 2026-08-06
fix(connectivity): dial ledger v3 gRPC by Service name, not TLS SNI

Dav-14:
ledgerAddress was built from backend.TLS.ServerName, the SNI/certificate FQDN
(ledger-<stack>.<stack>.svc.cluster.local). On clusters whose DNS domain is not
cluster.local that FQDN need not resolve, so connectivity-core could not reach
the ledger. Build the dial address from backend.Name — the Service name, which
resolves in the stack namespace on any DNS domain — and keep TLS.ServerName as
the SNI on spec.ledgerTLS for certificate verification, mirroring how the gateway
upstream dials its backend.

Addresses NumaryBot finding 52f0f4a930fa3d33.

---


Dav-14:
Superseded by #492, which includes this fix and verifies the Service dial address while preserving the TLS SNI.

